PR #38, opened 2015-02-14
Partial solution for UI refresh error

iasEnvy:
Celtic Minstrel and I figured out that windows created immediately after another don't have trouble refreshing or updating their content... so... it now creates a window, closes it before it has time to render, and then create the one we want. 

Please don't hit me.


CelticMinstrel:
Merged this manually.

PR #41, opened 2015-07-09
Master

Ircher:


CelticMinstrel:
The code in _Win32/_ is now frozen and is only kept around in case it still has useful things that can be pulled into mainline. All development (both for Mac and Windows) is now under _src/_. If you'd like to submit patches there, I'd be happy to consider them, but this one I'm going to close.

Ircher:
The Win32/ directory should be considered "frozen" from this point on. Apart from minor bugfixes, it should not be altered.

You don't consider this a bugfix?

CelticMinstrel:
I would consider it a bugfix. It's a fix for an area of the interface that no longer exists in the latest releases. I was considering a stronger statement than what you quoted; I don't see much point in even submitting bugfixes for the _Win32/_ legacy branch.

However, I'll bring this to the attention of @sylae - maybe she'll disagree with me and merge it.

sylae:
Personally, I question why we are even keeping around the legacy code? The only thing I am aware of that even touches it is the old Jenkins build script, which hopefully soon will be kicking the bucket. Perhaps it is in our interest to branch Win32/ out of master.

CelticMinstrel:
Really the only reason is "just in case there's something that didn't get merged in", plus the Jenkins build script.

PR #54, opened 2017-04-04
Small Update to Linux Branch

absquatulate:
Added X11 to scons build script
Added debug argument to scons script
Used external program 'zenity' to handle file save/load dialogs

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -50,6 +50,11 @@ if str(platform) == "posix":
 	env.Append(CXXFLAGS="-std=c++11 -stdlib=libstdc++")
 	env["CC"] = 'clang'
 	env["CXX"] = 'clang++'
+	env.Append(LIBPATH=Split("""
+		/usr/lib
+	"""), CPPPATH=Split("""
+		/usr/include
+	"""))
 if str(platform) == "darwin":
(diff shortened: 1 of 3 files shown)


CelticMinstrel:
Looks good! I hadn't actually considered using zenity for save/load dialogs because it's an external program, but I guess it's not a bad choice.

I notice you've commented out most of the cursor stuff, was that just because you couldn't get it to work?

absquatulate:
Yeah, it's lazy, I admit. I figure something is better than nothing until an actual GUI solution is implemented.

As for cursors, I commented it out just to get it to compile for now. I might look at implementing cursors soon, though they might have to be in only two colors.

CelticMinstrel:
Yeah, as far as I can tell, the X11 cursor routines only support monochrome cursors, which is rather unfortunate in my opinion. The simplest way to map colour cursors to monochrome would probably to only use the alpha channel, not sure if that would cause weird results though.

I guess I'll go ahead and merge this for now, anyway. If you have further work you can add another PR.

wrldwzrd89:
Does anyone know if Wayland or Mir fixes this Linux specific annoyance?

clort81:
Games can hide the X cursor and implement their own, getting pointer position from X.

PR #62, opened 2018-02-11
Fix broken links in README.md

wrldwzrd89:
2 links were affected by hard wrapping, unintentionally splitting them into two (broken) segments. This commit fixes that.


CelticMinstrel:
Thanks for noticing.

PR #63, opened 2018-02-13
Erase specials refactor

daerogami:
This cleans up the `erase_out_specials()` method in `boe.town.cpp` and introduces a few new helper methods.
Also introduces a minor fix on what is now `erase_completed_specials()` where we weren't ensuring the index did not exceed the vectors elements;

### DIFF
--- a/src/game/boe.town.cpp
+++ b/src/game/boe.town.cpp
@@ -21,13 +21,11 @@
 #include "boe.specials.hpp"
 #include "boe.infodlg.hpp"
 #include "mathutil.hpp"
-#include "boe.main.hpp"
 #include "render_image.hpp"
 #include "render_shapes.hpp"
 #include "render_text.hpp"
 #include "tiling.hpp"
 #include "strdlog.hpp"
-#include "fileio.hpp"
 #include "winutil.hpp"
 #include "res_image.hpp"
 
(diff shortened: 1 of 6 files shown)


src/game/boe.town.cpp
```diff
@@ -1247,52 +1245,79 @@ void erase_specials() {
...
+
+void erase_hidden_towns(cOutdoors& sector, int quadrant_x, int quadrant_y)
+{
+	for (short tile_index = 0; tile_index < sector.city_locs.size(); tile_index++) {
+		if (!sector.is_city_on_map(tile_index) ||
+			!is_tile_of_type(sector, tile_index, eTerSpec::TOWN_ENTRANCE) ||
+			!sector.is_tile_on_edge(tile_index)) {
+			continue;
+		}
+		if (!town_exists(sector, tile_index)) { continue; }
```
CelticMinstrel: If you want a one-liner if statement, please don't use braces. If you absolutely must use braces, please break it into three lines.

(My personal preference is the former, for the record.)

src/game/boe.town.cpp
```diff
@@ -1247,52 +1245,79 @@ void erase_specials() {
...
+				add_string_to_buf("Outdoor section corrupt. Problem fixed.");
+				sector.special_locs[tile_index].spec = -1; // TODO: Again, was there a reason for commenting this out?
 			}
+
+			sector.special_spot[where.x][where.y] = false;
+		}
+	}
+}
+
+cOutdoors& get_sector_ptr(int x, int y)
```
CelticMinstrel: This returns a reference, not a pointer, so please name it accordingly - either `get_sector_ref` or simply `get_sector`.

It could also be added to cUniverse or cScenario rather than being a global function, I guess... though maybe that slightly defeats the point?

src/game/boe.town.cpp
```diff
@@ -1247,52 +1245,79 @@ void erase_specials() {
...
+{
+	return *univ.scenario.outdoors[x][y];
+}
+
+bool is_tile_in_range(spec_loc_t tile, int max_range)
```
CelticMinstrel: I think this should probably be just `bool cArea::is_on_map(location tile)`. (Note that `spec_loc_t` is also a `location`, so it'll still work for your purposes.)

src/game/boe.town.cpp
```diff
@@ -1247,52 +1245,79 @@ void erase_specials() {
...
+{
+	return tile.x < max_range && tile.y < max_range && tile.x >= 0 && tile.y >= 0;
+}
+
+bool town_exists(cOutdoors& sector, short tile_index)
```
CelticMinstrel: From the name of this function, I would expect it to take only the index of the town (which is obtained from `spec` in this implementation). However... perhaps the better option here is to rename the function. Rather than simply checking whether a town exists (yes it's doing that, but there's a bit more to it), it's checking whether a town entrance leads to an existing town.

Also, its first argument is cOutdoors, so why not add it as a member of that class instead?

src/game/boe.town.cpp
```diff
@@ -1247,52 +1245,79 @@ void erase_specials() {
...
+{
+	auto towns_in_scenario = univ.scenario.towns.size();
+	return sector.city_locs[tile_index].spec >= 0 && sector.city_locs[tile_index].spec < towns_in_scenario;
+}
+
+bool is_tile_of_type(cOutdoors& sector, short tile_index, eTerSpec type)
```
CelticMinstrel: I see `type` as being a touch misleading in this function name. This is asking for the special property of a terrain space. Also, I think it's better for it to be taking a `location` rather than an index into `city_locs`.

Ideally this would go in `cArea`... that might require it to be virtual though, so maybe there are better options. That said, I'm not opposed to it being virtual.

src/scenario/outdoors.cpp
```diff
@@ -199,3 +200,21 @@ bool cOutdoors::cWandering::isNull(){
 void cOutdoors::reattach(cScenario& scen) {
 	scenario = &scen;
 }
+
+bool cOutdoors::is_city_on_map(short tile_index)
+{
+	return city_locs[tile_index].spec >= 0 && city_locs[tile_index].x < AREA_MEDIUM;
+}
+
+bool cOutdoors::is_tile_on_edge(short tile_index)
```
CelticMinstrel: I think this is better added to `cArea` and changed to take a `location` instead of an index into `city_locs`.

src/scenario/outdoors.cpp
```diff
@@ -199,3 +200,21 @@ bool cOutdoors::cWandering::isNull(){
...
+{
+	return city_locs[tile_index].x == minmax(0, AREA_MEDIUM - 1, city_locs[tile_index].x) &&
+		city_locs[tile_index].y == minmax(0, AREA_MEDIUM - 1, city_locs[tile_index].y);
+}
+
+unsigned short cOutdoors::get_terrain_index(short tile_index)
```
CelticMinstrel: This one... quite frankly it doesn't seem very useful, mainly because (once again) it takes an index into `city_locs` and then, besides that, it doesn't really do much of anything. Adding an `operator[]` in `vector2d` that takes `location` would probably be more useful (though I think I was purposely avoiding making `vector2d` depend on `location`...).

(That said, if kept you should be using the `ter_num_t` alias for the return type.)

src/game/boe.town.cpp
daerogami: Because it depends on accessing the universe instance ```auto towns_in_scenario = univ.scenario.towns.size();```

src/game/boe.town.hpp
```diff
@@ -21,6 +22,12 @@ void dump_gold(short print_mes);
 void pick_lock(location where,short pc_num);
 void bash_door(location where,short pc_num);
 void erase_specials();
+void erase_hidden_towns(cOutdoors& sector, int quadrant_x, int quadrant_y);
+void erase_completed_specials(cOutdoors& sector);
+
+/*NEW HELPERS, NEED TO GO SOMEWHERE*/
+bool does_location_have_special(cOutdoors& sector, location loc, eTerSpec type);
```
CelticMinstrel: Hmm, I like where you're going with this name, but it's a bit long (`location_has_special` maybe?) and there's a slight ambiguity about what "special" means (which may be unavoidable though, since it was already present in the underlying structures).

A logical place for this function is in cOutdoors, or possibly even cArea since it would also make sense calling it on a cTown.

src/scenario/area.hpp
```diff
@@ -41,6 +42,17 @@ class cArea {
 		, terrain(dim, dim)
 		, maps(dim, boost::dynamic_bitset<>(dim))
 	{}
+
+	bool is_on_map(location loc) const
+	{
+		return loc.x < max_dim && loc.y < max_dim && loc.x >= 0 && loc.y >= 0;
```
CelticMinstrel: Uh, wait a second, isn't this... a little backwards? I mean, it should be <0 and >=max_dim, right?

src/scenario/area.hpp
```diff
@@ -41,6 +42,17 @@ class cArea {
...
+	}
+
+	bool is_tile_on_edge(location loc) const
```
CelticMinstrel: Between this and the above function, I think their names should be consistent - either both include "tile" or neither does. (I tend to prefer shorter names, but I guess the difference in this case is fairly small.)

src/scenario/outdoors.cpp
```diff
@@ -199,3 +199,8 @@ bool cOutdoors::cWandering::isNull(){
 void cOutdoors::reattach(cScenario& scen) {
 	scenario = &scen;
 }
+
+bool cOutdoors::is_city_on_map(spec_loc_t loc)
```
CelticMinstrel: I don't understand, why are you checking `x` and not `y`? Shouldn't it be more like...

```
return loc.spec > 0 && is_on_map(loc);
```

src/scenario/scenario.cpp
```diff
@@ -528,3 +528,14 @@ cItem cScenario::return_treasure(int loot, bool allow_junk) {
 		treas.value = 0;
 	return treas;
 }
+
+cOutdoors& cScenario::get_sector(int x, int y)
+{
+	return *outdoors[x][y];
+}
+
+bool cScenario::town_exists(spec_loc_t loc) const
```
CelticMinstrel: Given that it takes a `spec_loc_t`, this is more "does the town entrance lead to a valid town" than "does this town exist" (which would just take the town index)...

I'm thinking maybe `city_valid` or something similar? Still not sure of the best name here.

This function could also be called with invalid data, since `spec_loc_t` is also used for things that have nothing to do with towns, but there's not a lot to be done about that (except maybe cloning the type for towns).

CelticMinstrel:
Doesn't look like you addressed anything mentioned in my second review?

daerogami:
I thought it was the same review, ill go through it again later this evening.

src/scenario/outdoors.cpp
daerogami: Because it is a direct refactor from ```sector.city_locs[k].spec >= 0 && sector.city_locs[k].x < 48``` on line 1261 on the commit preceding this branch
I'm not sure this is the right name for this method now that I read this, see comments on your review of is_tile_on_edge

src/scenario/area.hpp
daerogami: Doesn't this code check if the tile is on the map (within the boundaries specified by the minmax)? I think this name is wrong, let me know if you agree.

src/scenario/area.hpp
daerogami: ... no? wouldn't that be 'not_on_map'? Unless the dimension is negative and so are the coordinates

src/scenario/scenario.cpp
daerogami: will revise, i have an idea

src/scenario/area.hpp
CelticMinstrel: I look at it now and it seems correct after all. I'm not sure what I was thinking when I wrote the original comment.

src/scenario/area.hpp
CelticMinstrel: Hmm. Let me analyze the possible cases for `x`, supposing for simplicity that `max_dim = 8`...

* `x < 0`, ie it's not on the map
-- `minmax(0, 7, x)` yields 0, which is not equal to x
-- return false
* `x = 0`, ie it's on the edge
-- `minmax(0, 7, 0)` yields 0, which is equal to x
-- return true
* `x = 5`, ie it's in the middle somewhere
-- `minmax(0, 7, 5)` yields 5, which is equal to x
-- return true
* `x = 7`, ie it's on the edge
-- `minmax(0, 7, 7)` yields 7, which is equal to x
-- return true
* `x >= 8`, ie it's off the map
-- `minmax(0, 7, x)` yields 7, which is not equal to x
-- return false

...so it looks like you're correct and this formula is actually equivalent to the one above, so you can pretty much remove this and use `is_on_map` instead.

src/scenario/outdoors.cpp
CelticMinstrel: https://github.com/calref/cboe/blob/master/src/game/boe.town.cpp#L1261-L1264

If you look at those lines, I think you'll find the x<0 check is actually superfluous when combined with the minmax tests.

So I guess instead of having a function you could just directly test whatever.spec>=0 in the source function (ie, inline).

CelticMinstrel:
Not entirely happy with the length of the new name, but it could be worse, I guess. At least it's "only" four words. >_>

src/scenario/area.hpp
CelticMinstrel: (Assuming minmax's signature is minmax(min, max, val), but anything else wouldn't make sense in context.)

src/scenario/outdoors.cpp
CelticMinstrel: I mean I suppose you could add an `is_valid()` function to `spec_loc_t` that returns `spec >= 0`, but... that seems a bit much to me.

Still, I'm particularly not opposed to it, if you prefer that.

daerogami:
@CelticMinstrel Squashed refactor and kept fix as separate commit;

PR #176, opened 2018-02-25
Adding vs2017 project in new proj directory parallel to src

daerogami:
Tests and Install project are still WIP but shouldn't be a hold up to getting this merged.
I will move the vs2013 proj in another PR.
I am holding off on tests because I will work on MSTEST integration for both vs2013 and vs2017 in a different PR.
Install project in vs2017 is not a priority because as long as it still works in vs2013, @CelticMinstrel's workflow for creating releases will remain unaffected.

### DIFF
--- a/rsrc/menus/BladesOfExile.rc
+++ b/rsrc/menus/BladesOfExile.rc
@@ -32,9 +32,9 @@ LANGUAGE LANG_ENGLISH, SUBLANG_ENGLISH_US
 
 // Icon with lowest ID value placed first to ensure application icon
 // remains consistent on all systems.
-IDI_BLADESOFEXILE       ICON                    "../../../rsrc/icons/win/BOE Icon.ico"
-IDI_BOESAVE             ICON                    "../../../rsrc/icons/win/BoE Save.ico"
-IDI_BOESCEN             ICON                    "../../../rsrc/icons/win/BoE Scenario.ico"
+IDI_BLADESOFEXILE       ICON                    "BOE Icon.ico"
+IDI_BOESAVE             ICON                    "BoE Save.ico"
+IDI_BOESCEN             ICON                    "BoE Scenario.ico"
 
 /////////////////////////////////////////////////////////////////////////////
 //
--- a/src/fileio/debugout.win.cpp
+++ b/src/fileio/debugout.win.cpp
@@ -15,8 +15,8 @@ public:
 	}
 protected:
 	int sync() {
-		output_debug_string(str().c_str());
-		str(std::basic_string<CharT>()); // Clear the string buffer
+		output_debug_string(this->str().c_str());
+		this->str(std::basic_string<CharT>()); // Clear the string buffer
 		return 0;
 	}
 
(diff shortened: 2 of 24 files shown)


CelticMinstrel:
I'm not sure what to think of that last commit (tests). I think I'd kinda prefer if it could be done without the stdafx PCH, at least (though I guess it's not that big a deal). Also, I'd leave out the Tests namespace unless Microsoft actually requires it.

For the "your code here" you could either call the main() function or look at the [catchorg/Catch2](https://github.com/catchorg/Catch2) documentation about how to use Catch without a main(). (Though it looks like our included version is seriously out-of-date now, so I'm not sure if you can find relevant documentation there.)

rsrc/menus/BladesOfExile.rc
```diff
@@ -32,9 +32,9 @@ LANGUAGE LANG_ENGLISH, SUBLANG_ENGLISH_US
 
 // Icon with lowest ID value placed first to ensure application icon
 // remains consistent on all systems.
-IDI_BLADESOFEXILE       ICON                    "../../../rsrc/icons/win/BOE Icon.ico"
-IDI_BOESAVE             ICON                    "../../../rsrc/icons/win/BoE Save.ico"
-IDI_BOESCEN             ICON                    "../../../rsrc/icons/win/BoE Scenario.ico"
+IDI_BLADESOFEXILE       ICON                    "BOE Icon.ico"
+IDI_BOESAVE             ICON                    "BoE Save.ico"
+IDI_BOESCEN             ICON                    "BoE Scenario.ico"
```
daerogami: This _might_ break the vs2013 project, but it will need to be addressed with the one is moved anyways.

src/fileio/debugout.win.cpp
```diff
@@ -15,8 +15,8 @@ class basic_debugbuf : public std::basic_stringbuf<CharT, TraitsT> {
 	}
 protected:
 	int sync() {
-		output_debug_string(str().c_str());
-		str(std::basic_string<CharT>()); // Clear the string buffer
+		output_debug_string(this->str().c_str());
+		this->str(std::basic_string<CharT>()); // Clear the string buffer
```
daerogami: Really weird bug preventing me from compilation. Test this in xCode and let me know if it breaks anything.

CelticMinstrel: That's impossible, this code isn't even used in Xcode. So there's nothing to worry about.

src/game/boe.graphics.cpp
```diff
@@ -31,7 +31,9 @@
 #include "boe.menus.hpp"
 #include "winutil.hpp"
 #include "prefs.hpp"
+#ifndef MSBUILD_GITREV
 #include "gitrev.hpp"
+#endif
```
daerogami: Just a preprocessor that we set if MSBUILD is handling this. Added a NuGet package called "gitinfo" that makes git status related stuff available to preprocessor.

src/fileio/debugout.win.cpp
daerogami: Awesome!

daerogami:
@CelticMinstrel The tests project is partially implemented on purpose. Both vs2013 and vs2017 will get MSTEST integration. It will probably be a companion PR built on top of the vs2013 move to proj folder.

PR #185, opened 2018-03-23
Fix save game filename newline bug on Linux.

seisatsu:
Checks if there is a newline at the end of fname before examining the file extension. Otherwise in addition to the unwanted newline, the logic catches at ```fname.substr(dot) != ".exg"``` (now on line 404) and an additional ".exg" is added to the end of the filename.

### DIFF
--- a/.travis-install-deps.sh
+++ b/.travis-install-deps.sh
@@ -7,5 +7,5 @@ if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
 	brew install scons sfml
 else
 	sudo apt-get update -qq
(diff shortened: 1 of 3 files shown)


CelticMinstrel:
Thanks for trying to help, but I don't think this is the proper fix. If I recall correctly, the newline is due to improperly parsing zenity' output (which is why it's Linux-specific), so if that were fixed the newline would simply go away and we wouldn't need to worry about it in this part of the code. 

The code hooking into zenity should be in the winutil file if you want to take a look.

PR #190, opened 2018-04-05
SCons usability improvements and GCC support for posix

### LINKED ISSUE #184: Requiring Clang on Linux is unorthodox. Update SConstruct and test.
Currently the build procedure for cboe requires the Clang compiler on Linux. Although it works, this is unorthodox. A quick attempt to modify the SConstruct to use GCC/G++ instead causes SConstruct to mistakenly detect that the compiler doesn't work. Someone with SCons know-how should modify the SConstruct correctly and see if GCC/G++ run into errors while trying to compile the source, or if (ideally) no further changes are needed to support building on Linux with the GNU toolset.

akien-mga:
The first commit uses the SCons `Variables` object to register (and validate) command line arguments and generate a `scons --help` output. Using things like `BoolVariable` makes our life easier as you can specify `debug=1`, `debug=ON`, `debug=True`, `debug=yes`, etc., basically anything that would make sense for a bool, and it should work.

I tried to keep the arguments names already in use for compatibility, but `64bit` is not a valid identifier for SCons options (starts with a number), so I used a string enum `bits` instead.

That commit also adds support for reading common C and C++ configuration flags from the environment and/or the command line. Those are used by the second commit to remove the hardcoded `CXX=clang++` on Linux and let the user define it.

Fixes #184.

----

Note that GCC builds doesn't work out of the box for me with GCC 5.5.0. #189 is needed to fix a first compilation, and then I hit a second one for which I'll open a ticket (*Edit:* #191).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,24 +3,26 @@ language: cpp
 sudo: required
 services: ['docker']
 
-script: CXXFLAGS=-DBOOST_NO_CXX11_SCOPED_ENUMS scons
-
-compiler:
- - clang
+branches:
+ except:
+  - legacy-win32
 
 matrix:
  include:
   - os: linux
-    sudo: required
     dist: trusty
+    compiler: clang
+  - os: linux
+    dist: trusty
+    compiler: gcc
   - os: osx
+    compiler: clang
 
 install:
  - ./.travis-install-deps.sh
 
-branches:
- except:
-  - legacy-win32
+script:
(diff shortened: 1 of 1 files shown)


akien-mga:
Are Travis CI builds disabled for PRs? That's where they are most useful IMO, it would be nice to have them in this case to confirm that it still builds fine on CI before merging.

CelticMinstrel:
I don't remember if or why they were disabled, but I can look into enabling Travis builds for this PR.

CelticMinstrel:
I just enabled PR builds on Travis, but I think you'll need to push something new before it notices them.

PR #193, opened 2018-06-17
Fix the scons build on macOS High Sierra

wrldwzrd89:
This makes the project compile on macOS High Sierra (10.13), by doing the following:
* Works around the missing support for custom icons.
* Fixes the dylib references, so the built app launches.
* Builds in 64-bit mode for macOS by default.


wrldwzrd89:
For some reason the build fails with this error, but it's not fatal. Restarting the build once this occurs makes it finish and succeed. Any ideas are welcome.

CelticMinstrel:
Whoa, lots of stuff in here.

I'll try it and see if it works for me as well, though it's probably not a blocker if it doesn't since I use Xcode instead.

wrldwzrd89:
Yeah, the last 20 commits probably should have been squashed together, but... that's because I didn't know what I was doing XD

CelticMinstrel:
Sorry this is taking so long; I don't have the time during the week to dedicate to testing this, and all sorts of things can come up on weekends. For example, it's not too likely I'll get it done this weekend, because it's going to be really hot which really interferes with my ability to do programming stuff.

I _will_ look at this eventually though, so please just be patient.

wrldwzrd89:
It's OK - I'm in a time crunch myself, so I completely understand. (Quite honestly, I forgot this was still outstanding.)

CelticMinstrel:
Well, unsurprisingly it doesn't work on my computer - the fix dylibs stage has a lot of "file not found" errors. I'll fiddle with it a bit and see if I can figure out the problem.

CelticMinstrel:
Looking at it, I see basically two issues:

* Some builds of Boost have the `-mt` suffix on the libs while others don't. I think I built Boost myself though, so that might be why it doesn't have it. Still, it's probably a good idea to somehow check for this case.
* Lib names hard-code the version; we are using different versions of SFML, for example, so the build script needs to somehow intelligently substitute the actual version in use.

If you can find a way to deal with these, let me know. If not, I'll probably continue to fiddle with it and figure something else.

wrldwzrd89:
Yes, that's the biggest outstanding issue left. The `-mt` thing for Boost is easy to deal with. The version variance is harder, but still solvable... I think. I'll have to investigate what's needed to get the Xcode build system up to speed with the SCons changes in a separate PR... but that's for later. Right now, what matters is getting the library handling improved and making the build not fail so Travis CI is happy.

CelticMinstrel:
Sure, but my inability to build it is not great.

wrldwzrd89:
Closing this as Mojave makes fixing the issues that remain no longer possible.

CelticMinstrel:
Not sure what that means, but okay?

wrldwzrd89:
It means I'm both unwilling and unable to move this to the finish line.

PR #260, opened 2020-02-11
Fix unclickable shop info icons

### LINKED ISSUE #247: Information icons in shop not clickable
The following icons are not clickable.

clickable

Additionally, clicking around the icons and scrolling that UI seems to eventually produce temporary weird-looking glitches.

x-qq:
* Fixed unclickable shop icons by applying the same offset to event catching rects as is applied to the shop UI when it's being drawn in the main window.
* Removed strange adjustment of mouse click coords in shop UI by (-5, -5) px.
* Removed click_shop_rect function that was doing "click emulation" in shop UI. Rationale: 1) it was causing visual glitches, because 2) various widgets do not get to decide when .display() is called, since other drawables may not be ready
* Removed unneeded shop scrollbar redraw call which may improve performance of that UI a bit

Fixes #247

### DIFF
--- a/src/game/boe.actions.cpp
+++ b/src/game/boe.actions.cpp
@@ -121,11 +121,11 @@ static void advance_time(bool did_something, bool need_redraw, bool need_reprint
 // This is defined in pc.editors.cpp since the PC editor also uses it
 extern void edit_stuff_done();
 
-void init_screen_locs() {
-	rectangle startup_base = {281,1,329,302};
+static void init_shopping_rects() {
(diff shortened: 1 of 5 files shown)


CelticMinstrel:
You've removed `click_shop_rect` for some reason and now `draw_shop_graphics` is neverr called with `true`, which if I understand correctly means the pressed effect when clicking will not occur. Can you fix this, please? Putting `click_shop_rect` back where it was would be the easiest way, but whatever way you prefer is fine too.

PR #261, opened 2020-02-11
Fix debug mode (and other universe flags) being randomly set on upon load

### LINKED ISSUE #198: Special steps showing up in console
When nearing and encounter or special node the console is filled with "steps". While outdoors, the encounter node have to be stepped in multiple times before the actual encounter starts.

issue_steps

### LINKED ISSUE #200: Enemies do not move or attack and die with one hit
Enemies encountered in special encounters outdoors/wandering monsters stay passive in combat mode. They do not move, cast, shoot or anything. Only if characters is adjacent and moving away they do the "retaliation" attack. 

Enemies also die with one hit from anything. Attached a save file.
[blades-of-exile-team-save.zip](https://github.com/calref/cboe/files/4008392/blades-of-exile-team-save.zip)

x-qq:
Upon loading a save, a new instance of universe is default-constructed and then moved into existing universe. The problem is that several boolean flags in universe do not have initializers, so they sometimes end up with garbage values.

Fixes #198 
Fixes #200

### DIFF
--- a/src/universe/universe.hpp
+++ b/src/universe/universe.hpp
@@ -202,7 +202,9 @@ public:
 	cCurTown town;
 	cCurOut out;
 	fs::path file;
-	bool debug_mode, ghost_mode, node_step_through;
+	bool debug_mode { false };
+	bool ghost_mode { false };
+	bool node_step_through { false };
 	
 	void clear_stored_pcs();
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
I would've used = initializers but whatever, it doesn't matter. Merging.

Issue #254, opened 2020-02-06
Blinlock: Reward for clearing mines not received.


clort81:
Cleared the mines in Blinlock.  Mayor claims reward is awarded, but nothing appears in inventory or shows-up on game-world view.  

Tried with main character unencumbered (dropped armor first) but still no broadsword received.  Is this a bug?

Savefile attached.
[blinlock_reward.exg.gz](https://github.com/calref/cboe/files/4164320/blinlock_reward.exg.gz)

CelticMinstrel:
So just to clarify: you asked about iether "task" or "reward", and he responds something like "good news the spirit is dead" but you don't get a sword?

The reward should be a normal steel broadsword (which might show up as just an unidentified broadsword). If every pack was full (24 items), you would miss it, but it should still give even if it would put you over the weight limit, as long as there's space.

I'm pretty sure the log should say who exactly received the item; if it doesn't, then no-one did, probably. However, if that happened, asking him about "reward" should produce the same response and thus give another chance at the item.

x-qq:
Can reproduce on current master.

clort81:
Yes i asked about the reward and he claims to give it to me. 

ADos seemed to confirm a different reward not given in another game.  He also said my savefile resulted in town empty of characters. If i understood correctly.

CelticMinstrel:
Your savefile doesn't appear to contain the data that would result in an empty town, so I suspect ADoS may have been loading it with an older, more broken build. (I do see one value that looks suspiciously wrong in party.txt though - SHOPSTOCK is set to MAX_INT for some reason.)

NQNStudios:
A similar bug happened to me while playing "On a ship to Algiers". I opened the scenario in the editor and tracked down the node that isn't working. It's a "One-Time Give Item" node. Then I opened valleydy in the editor to compare, and found it is the same node type. Now I'll look into the code to see where it's going wrong and fix.

If possible, I will also contribute a unit test for this.

PR #277, opened 2020-04-27
try to correct some display's problems on OSX

fosnola:


CelticMinstrel:
Thank you for the contribution. Could you please explain:

* Why have you changed some cases of `setActive(true)` to `setActive(false)`?
* What's the reason for commenting out those two lines with `#if 0`?
* Why do you need a cast to `Uint8` for the colour?

Additionally, I can see in some places you seem to have replaced tabs with spaces, so if you could undo that (either revert and redo, or just push a fix commit), that would be great.

fosnola:
Hello,
- Why have you changed some cases of setActive(true) to setActive(false)?

	I do not know SMFL API and I tried to correct some display problems on OSX: the dialogs often flicker (between a different foreground/background bitmap), mini_map always flickers. 

So I looked for similar problems on the web, adding a glFlush seems to help (but not always and I don't like this solution), I also saw some references to setActive (so I tried to play with setActive). At least, on my MacBook, I no longer see any more flickering in the dialogs and the mini_map.

**Note**: I guess if you can draw in an inactive window and change its display (even if it's not really intuitive), it's better to do so. I also guess the problems I've seen, are related to that, somehow OSX makes some cache of the dialog/... bitmap when the redrawing is not complete... 

- What's the reason for commenting out those two lines with #if 0?

	Because when reading the comment, this seems a hack to correct the flickering problem on OSX ; if the solution is doing the drawing in a inactive windows, we probably no longer need this hack (so I _comment_ it).

- Why do you need a cast to Uint8 for the colour?

	Clang considers that converting these 3 int to 3 Uint8 is an error, ... It also considers that using '**<**' instead of '**"**' in gzstream.cpp is an error.

- Additionally, I can see in some places you seem to have replaced tabs with spaces, so if you could undo that (either revert and redo, or just push a fix commit), that would be great.

      I will do that latter.

fosnola:
So I have reverted and redone these two commits in the fork...

CelticMinstrel:
My understanding is that you can _only_ draw in the active window. If the window you draw to is not active, SFML will automatically set it active before drawing in it. The only reason I know of to manually `setActive(false)` are if you need to draw to the same window in two different threads, which should not be the case here.

So, if that's correct, then the `setActive(false)` is not doing anything useful, and removing it would have the same effect. (Unless deactivating the context forces some sort of cleanup work that I'm not aware of?)

The only other thing I'd add is to please use `uint8_t` instead of `Uint8` for the colour. They're technically the same thing, I just prefer the std spelling to the SFML spelling.

Since this branch is kinda messy, I probably won't directly merge it with the merge button, but will probably merge it manually instead, whenever I have time.

fosnola:
I have just commented my added _"setActivate(false);"_ to check if there are needed: the dialogs and the mini_map are flashing again :-~ 

**Notes**:
- my MacBook Air is very old: ~6 years old, so I guess it doesn't show up on a more modern computer,
- I have sometimes seen flickering in other places: in the inventory (when I want to identify/sell an object) and in talk windows; they are less disturbing and doing <<setActivate(false); draw data;setActivate()>> fix them. 
- there is already a _setActivate(false)_ presents in _cDialog::process_keystroke(cKey keyHit)_

> The only other thing I'd add is to please use uint8_t instead of Uint8 for the colour. They're technically the same thing, I just prefer the std spelling to the SFML spelling.

Ok.

> Since this branch is kinda messy, I probably won't directly merge it with the merge button, but will probably merge it manually instead, whenever I have time.

I can destroy the fork and recreate it, if you want.

**Note**:
- the last commit fixes the outdoor map problem ; but I do not know if it is better to call save_outdoor_maps when the file is saved, or when the file is loaded or in finish_load_party.

CelticMinstrel:
> I can destroy the fork and recreate it, if you want.

It's not really worth recreating the fork. If you want to clean stuff up and open a new pull request on a branch, that's fine; it probably won't make me get to it any sooner though. However, keep in mind that with the way you've opened this pull request, anything new you push will automatically be included.

For example, commits 602d9a4 and 833cca9… do those actually fix a visible issue, or did you only do that because it _looked_ wrong? Either way, I don't want to merge those commits in; but if there's a genuine issue in that area, I can probably find an alternate solution.

fosnola:
> It's not really worth recreating the fork. If you want to clean stuff up and open a new pull request on a branch, that's fine; it probably won't make me get to it any sooner though.

Ok.

> However, keep in mind that with the way you've opened this pull request, anything new you push will automatically be included.

Yes, I saw that :-~

**Note**: 
- there's still a few bugs I want to find, if I'm successful, there'll be more commits. The main thing is that using the menu to open a dialog makes the application crash :-~ 

> For example, commits 602d9a4 and 833cca9… do those actually fix a visible issue, or did you only do that because it looked wrong? Either way, I don't want to merge those commits in; but if there's a genuine issue in that area, I can probably find an alternate solution.

In these two commits, I want to correct some _problematic_ code, for instance,
```
bool cCurTown::is_explored(short x, short y) const{
 	if(x > record()->max_dim || y > record()->max_dim) return false;
 	return fields[x][y] & SPECIAL_EXPLORED;	
 }
```

makes no sense to me, because fields [x][y] is a unsigned long and SPECIAL_EXPLORED=0, i.e. it currently always returns false.

CelticMinstrel:
Hmm, looks like you're finding a bunch of memory issues. I did plan to eventually run asan on the game, so it's great to see this, but I'm not sure all of those are fixed in the correct way. If nothing else it's at least a guide I can use to find a fix, so thanks for doing it!

> In these two commits, I want to correct some problematic code, for instance,
> 
>  bool cCurTown::is_explored(short x, short y) const{
>  	if(x > record()->max_dim || y > record()->max_dim) return false;
>  	return fields[x][y] & SPECIAL_EXPLORED;	
>  }
> 
> makes no sense to me, because fields [x][y] is a unsigned long and SPECIAL_EXPLORED=0, > i.e. it currently always returns false.

You're missing a crucial detail here. It's not &'ing with the value of `SPECIAL_EXPLORED` - instead, it's calling the below overloaded operator (defined just below `eFieldType` in the same file), which means it's &'ing with 1, not 0. Thus it will return true if the least significant bit is set in `fields[x][y]`.

```
inline unsigned long operator&(unsigned long a, eFieldType b) {
	return a & (1 << b);
}
```

I understand that this can be confusing, and maybe there is a way to make it clearer that it's not doing what you thought - for example, if `field[x][y]` were a custom type instead of a bare integer. But unless _I'm_ missing something, there's no actual bug in my code here.

fosnola:
> Hmm, looks like you're finding a bunch of memory issues. I did plan to eventually run asan on the game, so it's great to see this, but I'm not sure all of those are fixed in the correct way. If nothing else it's at least a guide I can use to find a fix, so thanks for doing it!

Okay. Ie. I tried to fix them without changing the code too much, but sometimes you have to make more changes.
> You're missing a crucial detail here. It's not &'ing with the value of SPECIAL_EXPLORED - instead, it's calling the below overloaded operator
> ...
> I understand that this can be confusing, and maybe there is a way to make it clearer that it's not doing what you thought - for example, if field[x][y] were a custom type instead of a bare integer. But unless I'm missing something, there's no actual bug in my code here.

Oops, yes, this should work (even if it's not standard at all). I will revert these changes on my Mac and see if everything is still ok.

**Note** : regarding menus and dialogs on OSX, I found similar problems with sfml and wx, sfml and python on the web, so ...

CelticMinstrel:
Ohhh, nice catch in 2aabe65, I'm surprised no-one spotted that one before now.

fosnola:
> Ohhh, nice catch in 2aabe65, I'm surprised no-one spotted that one before now.

Not sure, but it might be able to solve https://github.com/calref/cboe/issues/254.

CelticMinstrel:
Just a minor thing in 23714adf177af48bc3ac33d8467f28222535f2e2: the outdoor sector (probably in univ.out->) _also_ has a `max_dim` value so it might be better to use that instead of hard-coding 96.

On 5995d67a8d516e205ab208658f67df6938b59627: probably the bug is that `caster` is bad, so I'll probably try and fix that instead of taking your commit.

fosnola:
> Just a minor thing in 23714ad: the outdoor sector (probably in univ.out->) also has a max_dim value so it might be better to use that instead of hard-coding 96.

Ok

> On 5995d67: probably the bug is that caster is bad, so I'll probably try and fix that instead of taking your commit.

This is the same bug as in https://github.com/fosnola/cboe/commit/ad1c0ee4758e9f5fe55a9040ce3abafcfd4af389, i.e. _caster_ comes from _cur_monst = &univ.town.monst[i]_. But when a monster is summoned, univ.town.monst has to get bigger, ...

clort81:
So cool.  Thank you for the work fosnola.  Hope that SSD is working out for you.

fosnola:
I don't know what _SSD_ means :-~ But, yes, my version is starting to work quite well, even if there are still some small display problems, ...

I'm testing scenarios to look for new bugs. Currently, _The Valley of Dying_, _A Small Rebellion_, _An Apology_, _Back to Normal_, _Nephil's Gambit_ seem to work, even though my playing style is bad (and I'm probably missing most of the side quests). I have to recheck The Riddle of the Spheres: I'm sure we can retrieve 5 of the 7 spheres, but my games have been screwed up by a bug that now must be fixed :-~

clort81:
Modified graphics assets on an ad-hoc basis might not be appropriate.  I have created a near total remaster of them and forked the project for this (Dark CBOE).

fosnola:
In fact, on a Mac retina, it is better to use a scale factor of 2 or 4; but if you use a scale factor of 4, images that are dithered look very bad. So, I just tried to replace some of them with undithered images, but I'm not an artist :-~ So, I hesitated to commit them or not and commit them to their home directory or another repository, ... but maybe, it can help someone who wants to play on a Mac (or not).

clort81:
It's not hard to swap out graphics. Maybe you have ideas how to make it easy for end-users to switch between alternates for the default (base) graphics?
Alternatively one could create high-res alternate campaign modules for the base campaigns.
Excited to review your work, but am juggling many things myself. Cheers.

fosnola:
It is probably possible to switch between alternate graphics in the application, but it doesn't seem so simple:

- I just had a quick look and SMFL doesn't seem to contain any code to choose a directory from a dialog; this means we have to add another library to do it or write platform-dependent code,
- switching graphics while the application is running is probably too error-prone, so we probably need to ask the user to restart the application,
- we also have to decide what happens if the user chooses a repository that does not contain all the necessary graphics (or some bad ones), reverts to the default repository and quits?

At the moment, I'm trying to _undither_ the images of the terrain. Then I will probably test if it is possible to use some high-resolution images.

fosnola:
I just reverted the previous commit and added in
- graphicsUndithered: the indexed images that have been undithered with https://github.com/kornelski/undither, it seems to improve these images a bit,
- graphicsUnditheredByHand: the pictures I had hand made.

fosnola:
To give some news, normally now the code must be able to read the resized images in the main graphics repository but not in the scenario file, i.e. gfxsheets.cpp still needs to be modified a lot. I have resized all the images in the main repository with ImageMagic: _convert the original .png -magnify final.png_ and I get (without magnification and with magnification) : Capture d’écran 2020-10-21 à 07 56 47
Capture d’écran 2020-10-21 à 07 53 56

CelticMinstrel:
I'm closing this because it's basically never going to get merged. It's not because I disagree with the changes but just because it's too huge and messy with everything muddled together (plus, I do disagree with _some_ of the changes, and disentangling will be a lot of work). In other words, I'm closing it because it's too big to review properly.

If you can break it down into simple, coherent chunks that do just one thing against current master (fix one issue, add one feature, etc), you can open separate pull requests for each one, and I will most likely accept them (though I may request some changes first). If you don't have time or motivation to do all that work (as there'd probably be some measure of redoing what you've already done), I will attempt to do it myself as I find the time. No promises on how long that will take, however.

I started trying to review it but gave up halfway, but here are a couple of comments I thought up while going through it.

* What is the purpose of the "graphicsUndithered" and "graphicsUnditheredByHand" folders? I'd prefer not to duplicate every graphic if that can be avoided. It'll make things complicated anytime I want to change graphics.
* The same question applies to the "scenarios_legacy" folder.
* I notice you disabled the "spell freebie" mechanic in legacy scenarios. This is not correct – it's an original feature of the game. I don't quite remember, but I believe it was used for spells cast by items… so that would mean using a spell item now costs extra AP in legacy scenarios, which is clearly wrong.
* "Found some unexistent monsters in the saved game, try to repair it." - Is that asking me to try and repair it or reporting that the game tried to repair it?
* I'd like a prose explanation of what the junk bag is and does.

fosnola:
(A quick answer )

_I started trying to review it but gave up halfway, but here are a couple of comments I thought up while going through it._

- _What is the purpose of the "graphicsUndithered" and "graphicsUnditheredByHand" folders? I'd prefer not to duplicate every graphic if that can be avoided. It'll make things complicated anytime I want to change graphics._

  The original graphics are dithered,  I try to "undither" some of them with imageMagick. I keep the original one in the graphics and store the other one elsewhere.  

- _The same question applies to the "scenarios_legacy" folder._

  This directory is not important, I used it to generate the final scenario (and to check if I didn't break some conversions).

- _I notice you disabled the "spell freebie" mechanic in legacy scenarios. This is not correct – it's an original feature of the game. I don't quite remember, but I believe it was used for spells cast by items… so that would mean using a spell item now costs extra AP in legacy scenarios, which is clearly wrong._

  If I remember correctly in the original game, the item casts a spell using the user's level (not the item's level). For example, in Death Valley, you find a lot of piercing crystals (level 1) that have almost no chance to dispel a barrier (if casted at level 1), .... I tried to allow these crystals to dispel barriers (even if it costs extra AP to an user).

- _"Found some unexistent monsters in the saved game, try to repair it." - Is that asking me to try and repair it or reporting that the game tried to repair it?_

  I wanted to say the saved game is broken. It would be better to use another saved game, even if you can continue with this save game.


- _I'd like a prose explanation of what the junk bag is and does._

I tried to implement a basic junk/trash bag (which is optional); the idea is that :
+ you can put item in the trash bag anywhere,
+ in a friendly city, you can retrieve/use/identify/sell the item in that trash bag. A friendly city is defined as a city where there are more friendly monsters than hostile monsters.
+ To avoid blockages, you can also retrieve an item in a city where it was put (or outside if it was put outside).

fosnola:
(regarding the junk bak which is a bit magical) In a friendly city, it interferes with the identification spell, which has only a 50% chance of identifying an item that is stored in the junk bak.

CelticMinstrel:
> * If I remember correctly in the original game, the item casts a spell using the user's level (not the item's level). For example, in Death Valley, you find a lot of piercing crystals (level 1) that have almost no chance to dispel a barrier (if casted at level 1), .... I tried to allow these crystals to dispel barriers (even if it costs extra AP to an user).

I don't understand how that's related to what I said about the "freebie" syntax. My point is that that's not a new feature; it was in the legacy game, so adding code to disable it on a legacy scenario is wrong. I don't quite remember what it was used for in the game; I'd have to dive through the code in some detail to figure it out. But it has always existed; the `spell_freebie` global is new, but the original game represented it by adding 1000 to the spell number.

fosnola:
Oops, I guess my problem was in fact with cast_town_spell in boe_party.cpp:
`
	short adj = spell_freebie ? 1 : univ.party[who_cast].stat_adj(eSkill::INTELLIGENCE);
        ...
		case eSpell::DISPEL_BARRIER:
			if((univ.town.is_fire_barr(where.x,where.y)) || (univ.town.is_force_barr(where.x,where.y))) {
				r1 = get_ran(1,1,100) - 5 * adj + 5 * (univ.town.difficulty / 10) + 25 * univ.town->strong_barriers;
`
whereas in the original sources, we find :
`
			case 41:
				if ((is_fire_barrier(where.x,where.y)) || (is_force_barrier(where.x,where.y))) {
						r1 = get_ran(1,0,100) - 5 * stat_adj(who_cast,2) + 5 * (c_town.difficulty / 10);
`
( I will have to revert this commit, probably this weekend ).

CelticMinstrel:
> catch.hpp has disappear the master branch of calref/cboe

Run `git submodule update --init`. (Or if you use a git UI, whatever the equivalent is.)

fosnola:
Ok.

fosnola:
**Notes:** 

- about redo/**fd334ecacf625c40a763a27d83ca6cd98730c82c** ( ~fosnola/master/be5e8e9df7ca8db51f48a1b90776b63de22d9ffc ):
I just added in this commit a new file _gfx/texture.hpp_ but I don't update any project because I modified my version of the OsX project to my liking, and I don't have MSVC...
- I did not include in redo, the commits that correspond to graphicUndithered, ... because if needed, they can be recovered directly

fosnola:
**Note:**

- about redo/e325d9bc I just added in this commit two new files game/minimap.{hpp,cpp} ...

PR #279, opened 2020-09-23
Restart Travis builds

madwizard:
Restart Travis-CI builds. To do this I had to:
- move SConstruct to python3. Minor changes, xrange to range and print "" to print()
- satisfy some compilers complaining about lack of includes
- disable tests... yep
- stop installing custom icons on Mac OS
- install TGUI as it's now dependency and Ubuntu doesn't seem to have it repos

### DIFF
--- a/.travis-install-tgui.sh
+++ b/.travis-install-tgui.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+set -ev
+
+env GIT_SSL_NO_VERIFY=true  git clone -b 0.8 --single-branch https://github.com/texus/TGUI.git
+cd TGUI && cmake . && make && sudo make install && cd /
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,30 @@
 
 language: cpp
-sudo: required
-services: ['docker']
-
 branches:
  except:
   - legacy-win32
 
-matrix:
+jobs:
  include:
   - os: linux
-    dist: trusty
+    dist: focal
     compiler: clang
   - os: linux
-    dist: trusty
+    dist: focal
     compiler: gcc
   - os: osx
-    osx_image: xcode6.4
+    osx_image: xcode9.3
     compiler: clang
+    env: BOOST_VERSION=1.73.0 BOOST_BUILD=false
+    addons:
+      homebrew:
+        packages:
+          - boost
+        update: false
 
-install:
+before_install:
  - ./.travis-install-deps.sh
+ - ./.travis-install-tgui.sh
 
 script:
  - scons CC=$CC CXX=$CXX CXXFLAGS=-DBOOST_NO_CXX11_SCOPED_ENUMS
@@ -28,7 +32,7 @@ script:
 notifications:
  email: false
  irc:
-  channels: ["irc.freenode.net#openboe"]
-  use_notice: true
-  on_success: change
-  on_failure: always
+        channels: [ "irc.freenode.net#wojslaw" ]
+        use_notice: true
+        on_success: change
+        on_failure: always
--- a/SConstruct
+++ b/SConstruct
@@ -34,14 +34,14 @@ env.Replace(tools=[toolset])
 
 # Check for platform support
 if platform not in ("darwin", "win32", "posix"):
-	print "Sorry, your platform is not supported."
-	print "Platform is:", platform
-	print "Specify OS=<your-platform> if you believe this is incorrect."
-	print "(Supported platforms are: darwin, win32, posix)"
+	print("Sorry, your platform is not supported.")
+	print("Platform is:", platform)
+	print("Specify OS=<your-platform> if you believe this is incorrect.")
+	print("(Supported platforms are: darwin, win32, posix)")
 	Exit(1)
-print 'Building for:', platform
-print 'Using toolchain:', toolset
-print 'C++ compiler:', env['CXX']
+print('Building for:', platform)
+print('Using toolchain:', toolset)
+print('C++ compiler:', env['CXX'])
 
 env.VariantDir('#build/obj', 'src')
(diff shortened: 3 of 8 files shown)


src/dialogxml/widgets/control.cpp
```diff
@@ -210,7 +210,7 @@ void cControl::setActive(bool active) {
 	depressed = active;
 }
 
-void cControl::setFormat(eFormat prop, short val) {
+void cControl::setFormat(eFormat prop, sf::Uint8 val) {
```
CelticMinstrel, line 213: I am a _little_ bit suspicious about this change to `sf::Uint8`. Are you certain that all existing uses fit into a single byte?

.travis-install-tgui.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -ev
```
CelticMinstrel, line 3: I kinda wonder why this needs to be a separate file instead of part of the main install file? I guess it doesn't hurt though.

.travis.yml
```diff
@@ -1,34 +1,38 @@
...
  - scons CC=$CC CXX=$CXX CXXFLAGS=-DBOOST_NO_CXX11_SCOPED_ENUMS
 
 notifications:
  email: false
  irc:
-  channels: ["irc.freenode.net#openboe"]
-  use_notice: true
-  on_success: change
-  on_failure: always
+        channels: [ "irc.freenode.net#wojslaw" ]
```
CelticMinstrel, line 35: This hunk will need to be reverted before I can merge the PR. Please revert both the channel change and the indentation changes.

SConstruct
```diff
@@ -52,14 +52,14 @@ if env['debug']:
...
-	fulltag = subprocess.check_output(["git", "tag", "--sort=v:refname"]).split('\n')[-1]
+	fulltag = subprocess.check_output(["git", "tag", "--sort=v:refname"]).decode('utf-8').split('\n')[-1]
 	tagrev = subprocess.check_output(["git", "rev-parse", fulltag]) if fulltag else ""
 	with open(target[0].path, 'w') as gitrev_hpp:
-		print >>gitrev_hpp
-		print >>gitrev_hpp, '#define GIT_REVISION "' + revid[0:7] + '"'
-		print >>gitrev_hpp, '#define GIT_TAG "' + fulltag + '"'
-		print >>gitrev_hpp, '#define GIT_TAG_REVISION "' + tagrev[0:7] + '"'
-		print >>gitrev_hpp
+		print(gitrev_hpp)
```
CelticMinstrel, line 58: I believe that this change may be incorrect? My understanding was that Python2 `print >>xxx` maps to Python3 `print(file=xxx)`.

SConstruct
```diff
@@ -182,6 +183,13 @@ elif platform == "win32":
 		env.Install(build_dir, source)
 elif platform == "posix":
 	env.Append(CXXFLAGS=["-std=c++14","-include","global.hpp"])
+	env.Append(LIBPATH=Split("""
```
CelticMinstrel, line 186: Please remove the excessive indentation here, so that the closing `"""` is at the same indentation level as `env.Append`.

rsrc/SConscript
```diff
@@ -77,30 +77,31 @@ if have_xmllint: # This is separate so that alternate xml validators could be us
...
-		'cursors/': 'boegraphics.icns',
-		'graphics/': 'boegraphics.icns',
-		'sounds/': 'boesounds.icns',
-	}
-	for dir, icon in icons.items():
-		env.SetDirIcon(
-			path.join("#build/Blades of Exile/data", dir, 'Icon\r'),
-			path.join("icons/mac/", icon)
-		)
+# if str(platform) == "darwin":
```
CelticMinstrel, line 83: Instead of commenting out this chunk, please enclose it in triple quotes (`"""` or `'''`).

CelticMinstrel (changes requested):
Looks good overall, just needs a few small changes and a double-check on the use of `sf::Uint8`.

src/dialogxml/widgets/control.cpp
madwizard: Hmm, seems like eControlType is outside the scope. Casting this short to sf::Uint8 will essentially do the same and cut overflowing values. I've made the change, because my gcc refused to compile it. Will check other options.

.travis-install-tgui.sh
madwizard: I wanted to keep tgui separate. In install-deps there are just packages, so I thought compilation is better kept in separate one.

.travis.yml
madwizard: I didn't want to spam #openboe with my trials, will revert.

SConstruct
madwizard: Yes, you are right, will fix.

src/dialogxml/widgets/control.cpp
madwizard: I added a cast to sf::Uint8, should be always in one byte range:

`newVal = sf::Color{(sf::Uint8)val, (sf::Uint8)val, (sf::Uint8)val};`

It compiles.

SConstruct
madwizard: Fixed

madwizard: Not really, it breaks Mac OS build.

CelticMinstrel:
Some general comments on your updates:

* The contents of those strings should also be less heavily indented.
* You need `file=gitrev_hpp` on the subsequent 4-5 lines as well.
* Casting definitely is safer than changing the type without being sure it's safe in all cases; I'd prefer a `static_cast` but am not heavily attached to that.

CelticMinstrel:
Closing since Travis is no longer in use, although if someone spots changes in this PR that are actually required, they are welcome to open a new PR with just those changes.

PR #283, opened 2022-03-13
make the game buildable on linux

x-qq:
Currently it does not build.

But with this patch it does.


src/skills_traits.hpp
```diff
@@ -9,6 +9,8 @@
 #ifndef BoE_SKILLS_TRAITS_HPP
 #define BoE_SKILLS_TRAITS_HPP
 
+#include <iostream>
```
CelticMinstrel, line 12: We don't want to include `<iostream>` in a header. Try `<iosfwd>` instead. The associated source file may need to add `<ostream>` and/or `<istream>`.

x-qq: good idea

PR #288, opened 2022-12-29
Update to TGUI 0.9.x

### LINKED ISSUE #284: Does not build with recent TGUI
TGUI 0.9 has introduced some breaking changes, our code will need to be adapted to them.

```
build/obj/game/boe.menu.cpp:200:52: error: cannot convert ‘const sf::String’ to ‘const tgui::String&’
  200 |         return this->tgui.get<tgui::MenuBar>(this->internal_menubar_widget_name);
      |                                              ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                    |
      |                                                    const sf::String
```

```
build/obj/game/boe.menu.cpp:238:30: error: ‘using element_type = class tgui::MenuBar’ {aka ‘class tgui::MenuBar’} has no member named ‘disconnect’
  238 |                 if(!menubar->disconnect(connection_id))
      |                              ^~~~~~~~~~
```

and possibly more.

NQNStudios:
These are the changes I had to make to compile against TGUI 0.9.1 (the version available in the Arch Linux user repository, AUR).

The main thing is that TGUI uses its own String class now instead of SFML's.

This should fix #284 .

I had to trigger the build with `scons INCLUDEPATH=$HOME/repos/cboe/deps/Catch2/include`. I thought it shouldn't be necessary to pass that INCLUDEPATH manually (and as an absolute path) when Catch2 is meant to be provided as a submodule by default, but that should probably be a separate issue.

Locally, the scons build ran without error and told me All tests passed. The game launches and I've created a party and opened a scenario successfully.

### DIFF
--- a/src/game/boe.menu.cpp
+++ b/src/game/boe.menu.cpp
@@ -235,7 +235,7 @@ void OpenBoEMenu::update_for_game_state(eGameMode overall_mode, bool party_in_me
 // Disconnect all spell menu items from signals and clear the menus
 void OpenBoEMenu::purge_spell_menus(tgui::MenuBar::Ptr& menubar) {
 	for(const auto& connection_id : this->spell_menus_connection_ids) {
-		if(!menubar->disconnect(connection_id))
+		if(!menubar->onMenuItemClick.disconnect(connection_id))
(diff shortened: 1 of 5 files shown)


CelticMinstrel:
Looks good… I don't really have any reason not to merge this.

Issue #208, opened 2020-01-04
Linux build: Main game window doesn't repair damage when Dialogue opened


clort81:
![damage](https://user-images.githubusercontent.com/51511737/71757709-07dc3480-2e98-11ea-8ac3-c441441a01a7.png)
The main game window doesn't repair damage (when stuff overwrites it) when Dialogue window is opened..

but it DOES... when map is invoked.   Maybe we can invoke the dialogue window the same way map window is?

CelticMinstrel:
We can't make the dialog windows behave the same as the map window, because the map window is a modeless window while the dialogs are modal.

From what I recall (haven't tested recently), this is not a problem on the Mac. I think it's because the Mac doesn't dirty the canvas of covered windows (or something). Anyway I don't know how this can be fixed while keeping the windows modal... there must be a way, but I don't know what it is.

NQNStudios:
I've just seen that this problem still affects the scenario editor--which makes sense, I was only trying to fix the game itself.

CelticMinstrel:
Presumably the same solution works in both editors though.

NQNStudios:
Yes, as long as they have complete void() draw functions, the editor main
functions can just pass them along the same way. I can look into it and
make another PR if you don’t get the chance.

NQNStudios:
We ended up determining that, due to SFML and operating system limitations, this can't be fixed further than it currently is by those PRs I made.

PR #289, opened 2023-01-04
call redraw_everything() when cDialogs move

NQNStudios:
Provides a partial fix for #208.

Based on this thread [here](https://en.sfml-dev.org/forums/index.php?topic=19658.0) it seems that SFML doesn't track an event for window movement. However, it tracks mouse movement and provides `RenderWindow.getPosition()`. So this patch tracks whether `win.getPosition()` reports a new position after a mouse movement event in cDialog. If the position has changed, it calls `redraw_everything()` and this fixes the artifacts left by dragging the dialog.

It's a little janky, but I've made a function pointer for `redraw_everything` in the cDialog class so that function can be provided in `boe.main.cpp` -- since it doesn't exist yet at cDialog's compile-time.

This is only a partial fix. It does not redraw over the artifacts left if the player drags the zenity file browser window, because that seems to be invoked as a subprocess and is not a cDialog.

If there are other dialogs that don't inherit cDialog, it also won't fix those cases.

And, as noted in the SFML forum thread, `getPosition()` won't report a window movement until the mouse has been released (very unfortunate). So the artifact will appear *while* you drag the dialogue, but at least it gets fixed when you finish dragging.

A real ideal solution would be if SFML provides a way to lock the buffer of the main window whenever a cDialog opens or the zenity subprocess is called. I haven't found such a feature yet, and wanted to share this partial fix as one option.

### DIFF
--- a/src/dialogxml/dialogs/dialog.cpp
+++ b/src/dialogxml/dialogs/dialog.cpp
@@ -35,6 +35,7 @@ extern sf::Texture bg_gworld;
 const short cDialog::BG_DARK = 5, cDialog::BG_LIGHT = 16;
 short cDialog::defaultBackground = cDialog::BG_DARK;
 cDialog* cDialog::topWindow = nullptr;
+void (*cDialog::redraw_everything)() = nullptr;
 
 std::string cDialog::generateRandomString(){
 	// Not bothering to seed, because it doesn't actually matter if it's truly random.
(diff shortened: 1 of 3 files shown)


CelticMinstrel:
It's kind of annoying that MacOS is smart enough not to have to do this but Windows and Linux are not, but I guess if it works, then that should be fine.

I'll do a quick check to make sure it builds on my computer and then probably merge it.

> If there are other dialogs that don't inherit cDialog, it also won't fix those cases.

I think that might mean the minimap window?

NQNStudios:
In testing, I’ve found that the minimap window doesn’t cause the artifact.
Someone said something in the #208 discussion about this problem only
affecting modal dialogs and that the map isn’t one. Could that be it?

I realized another solution that would handle all cases, and for most
(maybe even all) modern games things are done this way: simply re-rendering
EVERY frame.

CelticMinstrel:
> Could that be it?

Certainly.

> simply re-rendering EVERY frame.

It wouldn't even need to fully re-render every frame – the terrain and most of the text-heavy areas are cached on off-screen surfaces, which there should be no need to keep updating when nothing has changed.

We do have a draw manager now. Ultimately I'd like it to be responsible for the entirety of the final phase of rendering. At that point it would be relatively easy to just call `draw_all()` on the main loop. But currently almost nothing is actually using the draw manager, so that's a lot of work.

PR #290, opened 2023-01-05
Enable CI using GitHub Actions

CelticMinstrel:
This is a pull request so I can more easily sort through it and figure out how things should be squashed together, because while the bulk of it could be compressed, there are at least a few things that definitely shouldn't be.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,201 @@
+{
+	name: CI,
+
+	on: {
+		push: {
+			branches: [ master, ci ]
+		},
+		pull_request: {
+			branches: [ master ]
+		}
+	},
+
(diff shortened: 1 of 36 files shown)


NQNStudios:
Apparently GitHub actions can upload automatic releases. This action [here](https://github.com/marketplace/actions/create-release) seems like a popular way to do so. Maybe uploading "nightly" builds could become part of the CI.

CelticMinstrel:
It's a good idea, but at this point I'd rather get the CI to even just function. Uploading artifacts can come once it works.

This _was_ working back when I worked on it. It likely doesn't work anymore due to dropping legacy systems, but we'll see…

CelticMinstrel:
Merged manually.

Issue #273, opened 2020-03-16
Required attributes missing on elements in multiple dialog xml files


seisatsu:
In the following xml files, the element "led" does not have the required attribute "width", which will likely lead to a crash when they are loaded. This has been confirmed with tip-of-the-day.xml and pc-alchemy-info.xml.
* edit-item-shortcut.xml
* edit-outdoor-encounter.xml
* edit-special-item.xml
* edit-town-details.xml
* edit-townperson.xml
* graphic-types.xml
* make-scenario1.xml
* make-scenario2.xml
* new-town.xml
* pc-alchemy-info.xml
* pc-spell-info.xml
* tip-of-day.xml

In addition, edit-party.xml throws `Required attribute anchor missing on element button (file edit-party.xml, line 4, column 5)`. This can be fixed by changing the "relative" attribute from "neg pos" to "neg abs".

I will continue appending to this issue as errors occur, while applying temporary fixes to my own files. (I do not know the appropriate widths for the led elements, but having something there will at least allow the game to continue on my end until fixes are pushed.)

CelticMinstrel:
I believe any width should fix the led elements, if it's less than 14 or so (the actual width of the button) it'll be expanded to that size.

I'm guessing a lot of these are older LEDs that haven't been updated to contain their label.

NQNStudios:
A crash in the scenario editor when I click Edit Town Details is preventing me from diagnosing a different bug, so I think I'll try to fix this.

It most be because `edit-town-details.xml` is on that list, because the missing attribute 'width' message is the one I'm getting.

CelticMinstrel:
Generally speaking, the appropriate width is "wide enough that the full label can fit" and also usually "same width as other LEDs in the same column".

CelticMinstrel:
By the way, you might want to check out #274 before getting started on this

CelticMinstrel:
I think this should be considered fixed by b03a200f8a54ff68bb79e8778c004dcb880d887d and c3dd886783c3fea2fb0f95c228e74b1b79827b5c?

NQNStudios:
With #298 in place, I agree.

PR #298, opened 2023-01-06
Test case for dialog XMLs

NQNStudios:
This test case should catch potential regressions of #273.

cDialog prepends data/dialogs onto the path you pass it, so I had to add an optional flag to not do that, so the tests can pass a path relative to their cwd.

### DIFF
--- a/src/fileio/resmgr/res_dialog.cpp
+++ b/src/fileio/resmgr/res_dialog.cpp
@@ -16,15 +16,7 @@ extern std::ostream& std_fmterr(std::ostream& out);
 class DialogLoader : public ResMgr::cLoader<DialogDefn> {
 	/// Load a dialog definition from an XML file.
 	DialogDefn* operator() (const fs::path& fpath) const override {
-		TiXmlBase::SetCondenseWhiteSpace(false);
-		ticpp::Document xml(fpath.string().c_str());
-		try {
-			xml.LoadFile();
-		} catch(ticpp::Exception& e) {
-			std::cerr << "Error reading XML file: " << e.what();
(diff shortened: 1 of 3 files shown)


CelticMinstrel:
It's hard to say because the log is extremely sparse, but I think the Win64 test is failing because the MSVC build has a different directory structure.

It looks like this prints the absolute path to the dialog when it fails. Would it not make more sense to print just the filename?

CelticMinstrel:
I went ahead and made one of the changes that I'd been planning for awhile, which renders your `prepend_path` obsolete. Instead you can basically do what the `DialogLoader` does – load the XML file and pass it to the `cDialog` constructor. (There's no reason to use the resource manager in the tests, I think.)

I'm not quite sure, but I think the directory structure in Windows should look something like this:
```
rsrc/dialogs/
proj/vs201x
    Tests/x64/Debug/
        Test.exe
    Game/x64/Debug/
        data/dialogs/
```

It's probably easier to grab dialogs from the repository root than the game build directory… the only complicated part is how to do it only when building with MSBuild. Checking `_WIN32` is no good, and I'm pretty sure `_MSC_VER` doesn't work either because the scons build also uses Microsoft Cl as the compiler. Maybe the only option is to add a preprocessor definition to Test.vcxproj…?

NQNStudios:
I rebased onto the DialogDefn changes, but have not fixed the Win64 test failure yet. 

I refactored the DialogLoader `operator()` body into a global function, `load_dialog_defn`, because I want to make sure the test is constructing the DialogDefn the same way the game does. I could have duplicated the code loading the XML node, but then the test code could get out of sync with the specific XML settings used in `res_dialog.cpp`.

PR #302, opened 2023-01-09
Linux handle zenity non-zero code. Close #301

### LINKED ISSUE #301: Clicking "Cancel" in Zenity file-picker on Linux causes core dump.
On Linux, if you click "Cancel" from a file picker dialog, you get a crash with a core dump (In the game, and both editors)

It's because here:
https://github.com/calref/cboe/blob/master/src/tools/winutil.linux.cpp#L134

There is no handling for a popen() that doesn't result in success.

I think I can fix this.

A related problem is that the Zenity picker doesn't let you choose a file that doesn't exist yet, so creating a new scenario in the editor is impossible (maybe you could do it by picking a file that DOES exist an overwriting it, but... that would be bad...)

NQNStudios:
I started trying to change the return type to boost::optional, but that turned out to be a lot of work and I realized the if statements would then have to check the optional for existence AND check the value for non-emptiness, which would just be silly.

### DIFF
--- a/src/tools/winutil.linux.cpp
+++ b/src/tools/winutil.linux.cpp
@@ -145,12 +145,17 @@ static std::string runFileDialog(const std::string& file, bool save) {
         if (fgets(buffer, 128, hfile) != NULL)
             filename += buffer;
     }
+	
+	// When the dialogue is canceled, return empty string.
+    int result = pclose(hfile);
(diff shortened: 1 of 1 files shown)


NQNStudios:
Looks like the checks failed because of a dependency installation getting a network error

PR #303, opened 2023-01-09
Windows file dialog return empty on cancel/error

### LINKED ISSUE #268: Cancelling a save still saves
I believe this refers to clicking "Cancel" in the actual save dialog.

[Reported by Thaluikhain](http://spiderwebforums.ipbhost.com/topic/25394-experimental-releases/?do=findComment&comment=309718)

NQNStudios:
If it compiles, this should fix #268.

According to this documentation:

https://learn.microsoft.com/en-us/windows/win32/api/commdlg/nf-commdlg-getopenfilenamea
https://learn.microsoft.com/en-us/windows/win32/api/commdlg/nf-commdlg-getsavefilenamea
https://learn.microsoft.com/en-us/windows/win32/api/commdlg/nf-commdlg-commdlgextendederror

When Get{Open/Save}FileNameA() finishes with the Cancel input, the function returns false and the error code is 0.

So the code had 2 problems before: 1) not checking for the cancel code 0, and 2) not returning an empty string on error.

Someone needs to test this on Windows

### DIFF
--- a/src/tools/winutil.win.cpp
+++ b/src/tools/winutil.win.cpp
@@ -208,8 +208,10 @@ static std::string runFileDlog(OPENFILENAMEA& dlg, const std::string& file, bool
 	if(save) err = GetSaveFileNameA(&dlg);
 	else err = GetOpenFileNameA(&dlg);
 	if(err == 0) {
-#define CASE(x) case x: std::cerr << "File dialog failed: " #x << std::endl; break
+#define CASE(x) case x: std::cerr << "File dialog failed: " #x << std::endl; return ""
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
Ugh, why is that function using manual memory management!? It looks like this change will leak the temporary buffer used for the path.

And that's just one of the problems. It should probably be migrated to Get{Open,Save}FileNameW() instead to properly handle Unicode paths… but that's out of scope for this now.

NQNStudios:
Closing in favor of #305.

PR #305, opened 2023-01-10
Make sure lpstrFile is empty on cancel

### LINKED ISSUE #268: Cancelling a save still saves

NQNStudios:
I built the game on Windows and found that I actually couldn't reproduce #268 -- because it seems that dlg.lpstrFile is actually an empty string in the case of a cancel or an error. At least, when built with Visual Studio 2017, that was the case when I debugged it a few times.

Maybe it's not guaranteed that lpstrFile will be an empty string, and random behavior caused this bug. I've added a commit that sets it to an empty string, which will guarantee that doesn't happen.

This, unlike my previous PR, also avoids leaking memory by following the same control flow as before.

This should close #268.

### DIFF
--- a/src/tools/winutil.win.cpp
+++ b/src/tools/winutil.win.cpp
@@ -208,6 +208,7 @@ static std::string runFileDlog(OPENFILENAMEA& dlg, const std::string& file, bool
 	if(save) err = GetSaveFileNameA(&dlg);
 	else err = GetOpenFileNameA(&dlg);
 	if(err == 0) {
+		dlg.lpstrFile = "";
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
The approach looks fine but the CI says it doesn't compile using the Microsoft toolchain. Perhaps setting the first character to 0 instead would work?

PR #307, opened 2023-01-13
Redo2

fosnola:
So a new pull request to replace https://github.com/calref/cboe/pull/285 ( after doing a rebase with master on 01/11/23)


CelticMinstrel:
Please delete the first commit and force-push.

fosnola:
> Please delete the first commit (my commit) and force-push.

This one: https://github.com/calref/cboe/pull/307/commits/97b5b6745d314e281c4511b9ec059de2433e1aff ? 

Ie. if I do that, 
```
@@ -92,6 +92,7 @@ bool mod_contains(eKeyMod mods, eKeyMod mod) {
 }

 unsigned char applyShift(unsigned char c){
 	if(c < ' ' || c >= 0x7f) return c;
 	static const char afterShift[] = {
 		' ', '!', '"', '#', '$', '%', '&', '"', '(', ')', '*', '+', '<', '_', '>', '?',
 		')', '!', '@', '#', '$', '%', '^', '&', '*', '(', ':', ':', '<', '+', '>', '?',
 @@ -104,6 +105,7 @@ unsigned char applyShift(unsigned char c){
 }

 unsigned char removeShift(unsigned char c){
 	if(c < ' ' || c >= 0x7f) return c;
 	static const char afterUnShift[] = {
 		' ', '1', '\'','3', '4', '5', '7', '\'','9', '0', '8', '=', ',', '-', '.', '/',
 		'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', ';', ';', ',', '=', '.', '/',
```
will disappear completely, i.e. git rebase deleted this part in my commit https://github.com/calref/cboe/pull/307/commits/c2afca92997cc4b6b43c60c460dfae584dad7315 ( because it was in conflict with master )

CelticMinstrel:
Okay, I'm going to slowly work my way through these commits, and in some cases I will open up a comment thread on a specific commit if I think it looks wrong or otherwise require additional explanation as to its purpose.

src/game/boe.fileio.cpp
```diff
@@ -93,6 +93,9 @@ void finish_load_party(){
 	
 	overall_mode = town_restore ? MODE_TOWN : MODE_OUTDOORS;
 	stat_screen_mode = MODE_INVEN;
+	// build_outdoors call add_outdoor_maps, so we must update them first
+	// checkme: true for actual files, unsure for a legacy saved file
+	save_outdoor_maps();
```
CelticMinstrel, line 108: I can't imagine any way that this would be correct. In simplified pseudo-code, what this does is:

```cpp
univ.scenario.outdoors.maps = univ.out.out_e;
univ.out.out_e = univ.scenario.outdoors.maps;
```

That clearly doesn't make any sense at all – it's putting garbage in `out_e` and then loading that garbage back in.

If the issue is that `out_e` contained garbage to begin with, then the solution would be to make it not contain garbage.

Thinking on it, I think the correct answer might ultimately be that `out_e` is just obsolete and we delete it entirely. I think its reason for existing was that the original game had such a tiny memory footprint that it couldn't afford to keep the entire scenario in memory all at the same time.

src/dialogxml/keycodes.cpp
```diff
@@ -100,6 +100,10 @@ unsigned char applyShift(unsigned char c){
 		'~', 'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O',
 		'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z', '{', '|', '}', '~',
 	};
+	if (c<' ')
+		return ' ';
+	if (c>=0x7f)
+		return c;
```
CelticMinstrel, line 106: Just noting that I already applied this fix myself, in a slightly different way.

src/game/boe.specials.cpp
```diff
@@ -2090,6 +2090,10 @@ cSpecial get_node(spec_num_t cur_spec, eSpecCtxType cur_spec_type) {
 			}
 			return univ.out->specials[cur_spec];
 		case eSpecCtxType::TOWN:
+			if (is_out()) {
+				showError("The scenario called a town special node but it is not in town.");
+				return dummy_node;
```
CelticMinstrel: What issue does this fix? I get that you're not supposed to call town specials while outdoors, but given that the commit message says "asan", I'm guessing you had a memory-related motive for enforcing it.

Is there any reason why the reverse is not also forbidden? You've forbidden calling a town special while outdoors, but not an outdoors special while in town, which seems odd.

Also, is there any chance that this causes special nodes in outdoors combat to fail? Testing outdoor encounters that call nodes on meet, win, or flee needs to be taken care of, obviously, but if I recall correctly, I also introduced a way to load an arbitrary town as your outdoor arena. I don't remember whether special encounters placed in that town will trigger, but if so, we would need to make sure that they work too.

fosnola: > What issue does this fix? I get that you're not supposed to call town specials while outdoors, but given that the commit message says "asan", I'm guessing you had a memory-related motive for enforcing it.
> 
> Is there any reason why the reverse is not also forbidden? You've forbidden calling a town special while outdoors, but not an outdoors special while in town, which seems odd.
> 
> Also, is there any chance that this causes special nodes in outdoors combat to fail? Testing outdoor encounters that call nodes on meet, win, or flee needs to be taken care of, obviously, but if I recall correctly, I also introduced a way to load an arbitrary town as your outdoor arena. I don't remember whether special encounters placed in that town will trigger, but if so, we would need to make sure that they work too.

This is from https://github.com/fosnola/cboe/commit/1e791f575e5dff039554ade1c89b390341a0e5f5 , so for some reason, get_node was called with a town special when I was out. But this is probably a quick fix for another problem that was fixed afterwards (maybe a city event that was stacked and called after calling end_town_mode).

We can probably keep this patch as a sanity check ( and add a similar patch when we are outdoor).

src/game/boe.fileio.cpp
fosnola: Yes this is tricky, I think that when we retrieved a save ( and if we are outdoor) the current map is retrieved by 
```
void cCurOut::readFrom(std::istream& file) {
	readArray(file, expl, 96, 96);
	readArray(file, out, 96, 96);
	readArray(file, out_e, 96, 96);
}
```
so if we do not want to loose it, we need first to retrieve the data of out_e, ... ( which was what save_outdoor_maps do ).

**Note** we need these two maps because originally, when we change sectors, the game rereads the files of the new sector (because it doesn't keep in memory all the maps of the sectors). A similar problem appears with the city, when we enter a city, normally we have to reread the city file (which we don't do). In my new branch I tried to correct it with https://github.com/fosnola/cboe/commit/684929456788b9cba4f90faf98fe7cf83b7a4eeb (and I also add a persistent flag so the coder can decide to have persistent town or town which are reset each time a user enter in a town )

CelticMinstrel: > we need two maps because originally, when we change sectors, the game rereads in the scenario the data corresponding to the new sector (because it doesn't keep in memory all the maps of the sectors). A similar problem appears with the city, when we enter a city, normally we have to reread the city file (which we don't do).

Right, but from what I understand, this was just a performance optimization and isn't needed. We could easily save maps for the entire scenario without issue.



> In my new branch I tried to correct it with [fosnola@6849294](https://github.com/fosnola/cboe/commit/684929456788b9cba4f90faf98fe7cf83b7a4eeb) (and I also add a persistent flag so the coder can decide to have persistent town or town which are reset each time a user enter in a town )

That commit doesn't really make much sense. It seems to be overly complicated for what sounds like a very simple feature.

src/game/boe.specials.cpp
CelticMinstrel: Keeping this patch (and a similar patch for outdoors) is fine as long as it doesn't cause any problems with specials in outdoor combat.

fosnola: I am not sure that I understand. If we are in outdoor combat, overall_mode is equal to MODE_COMBAT ? So is_out() will return false:
```
bool is_out() {
        if((overall_mode == MODE_OUTDOORS) || (overall_mode == MODE_LOOK_OUTDOORS))
                return true;
        else if(overall_mode == MODE_SHOPPING) {
                std::swap(overall_mode, store_pre_shop_mode);
                bool ret = is_out();
                std::swap(overall_mode, store_pre_shop_mode);
                return ret;
        } else if(overall_mode == MODE_TALKING) {
                std::swap(overall_mode, store_pre_talk_mode);
                bool ret = is_out();
                std::swap(overall_mode, store_pre_talk_mode);
                return ret;
        } else return false;
}
```
Ie. I don't think it's possible to shop or talk when we're in external combat, but maybe I'm wrong...

CelticMinstrel: Ah, so I think that means there should be no issue either with this or with a similar check in the outdoors case. The on meet special is called before combat starts, and on flee/win is called after it ends.

src/game/boe.newgraph.cpp
```diff
@@ -869,8 +869,14 @@ std::string get_item_interesting_string(cItem item) {
 		case eItemType::HELM:
 		case eItemType::GLOVES:
 		case eItemType::SHIELD_2:
-		case eItemType::BOOTS: // TODO: Verify that this is displayed correctly
-			sout << "Blocks " << item.item_level + ((item.protection > 0) ? 1 : 0) << '-' << item.item_level + item.protection << " damage";
+		case eItemType::BOOTS: // coherent with damage_pc
+			sout << "Blocks "
+			<< (item.item_level>0 ? 1 : item.item_level) + (item.bonus>0 ? 1+item.bonus/2 : item.bonus) +
+				(item.protection > 0 ? 1 : item.protection)
```
CelticMinstrel: This new formula looks extremely weird. Can you point to the exact lines of code that it's "coherent" with?

In particular, the `item_level>0 ? 1 : item_level` bit doesn't seem right.

src/game/boe.specials.cpp
```diff
@@ -1855,7 +1855,7 @@ void special_increase_age(long length, bool queue) {
 			if(univ.town->timers[i].time > 0) {
 				short time = univ.town->timers[i].time;
 				bool need_redraw = false;
-				for(unsigned long j = age_before + (time == 1); j <= current_age; j++)
+				for(unsigned long j = age_before + 1; j <= current_age; j++)
```
CelticMinstrel, line 1874: So, if I understand correctly, the bug is that a timer could be called on the turn after it was called? And this definitely fixes it? Just double-checking since the commit message contains "try".

As an aside: the original source code looks dramatically different so it's hard to compare, but setting the timer's time to 0 should also prevent it from double-triggering.

CelticMinstrel:
General note: for the asan commits I'm going to be investigating and doing my own thing, which may in some cases be different from what you did. So, when you next rebase this PR, the asan commits will probably conflict, but in most cases you can just skip them. (Do double-check to make sure I didn't miss something though.)

src/tools/vector2d.hpp
```diff
@@ -152,7 +152,8 @@ template<typename Type, typename Alloc = std::allocator<Type>> class vector2d {
 		size_t old_w = w, old_h = h;
 		w = width; h = height;
 		data.resize(w * h);
-		if(old_w < width) {
+		// ASAN undefined behaviour if old_h==0, y=old_h-1 is ...
+		if(old_w < width && old_h) {
```
CelticMinstrel: I've rewritten this function with unit tests, so presumably this change doesn't apply anymore?

src/scenario/scenario.cpp
```diff
@@ -249,7 +249,14 @@ void cScenario::import_legacy(legacy::scenario_data_type& old){
 	horses.resize(30);
 	for(short i = 0; i < 30; i++) {
 		boats[i].import_legacy(old.scen_boats[i]);
+		// in the .exs scenario, the boats in town seem to be defined with exists=false
```
CelticMinstrel, line 395: This seems suspicious. What's the bug that this is even trying to fix? I think the issue might be rather that `exists` isn't relevant in a scenario file and is supposed to be overwritten when the scenario is loaded for a party, but I'm not sure of the details.

This could also possibly be a symptom of an issue with the legacy structs? They're supposed to be defined without padding, but is it possible the compiler is ignoring the pack pragma?

src/game/boe.locutils.cpp
```diff
@@ -250,20 +250,17 @@ void update_explored(location dest) {
...
+				}
 			}
 	} else {
 		make_explored(dest.x,dest.y);
 		for(look2.x = max(0,dest.x - 4); look2.x < min(univ.town->max_dim,dest.x + 5); look2.x++)
 			for(look2.y = max(0,dest.y - 4); look2.y < min(univ.town->max_dim,dest.y + 5); look2.y++)
-				if(!is_explored(look2.x,look2.y))
-					if((can_see_light(dest, look2,sight_obscurity) < 5) && (pt_in_light(dest,look2)))
-						make_explored(look2.x,look2.y);
+				if(!is_explored(look2.x,look2.y) && can_see_light(dest, look2,sight_obscurity) < 5 && pt_in_light(dest,look2))
```
CelticMinstrel, line 245: Did this fix an actual issue? It looks like just merging two if statements that didn't really need to be separate.

src/game/boe.specials.cpp
fosnola: Yes, I tried to avoid double triggering (which could appear for this reason or another) ; the previous code was not logical, ie. when length=1 and time!=1, the execution of the code was:
```
for(unsigned long j = party.age-1 ; j <= party.age; j++) ...
party.age++
for(unsigned long j = party.age-1 ; j <= party.age; j++) ...
```
( ie. we tried to trigger each timer twice )

src/scenario/scenario.cpp
fosnola: Yes, you can probably skip this patch (ie. _exists_ was not relevant in a scenario file, but there exists some code in start_town_mode which does that).

**Notes**:  I guess I was trying to figure out why some boats/horses were disappearing outside but there is another patch to fix that.

src/game/boe.locutils.cpp
fosnola: This is probably the merging of two if statements into one. 

I started to simplify the code a bit before trying to get a code that worked a bit better, but I needed a lot of trials. So I had commit this simplification at one point and later when I had something that worked better, the big changes.

fosnola:
For your information, in Xcode, you can easily add an asan check by using the menu Product, Scheme,  Edit Scheme... and then clicking on Run Debug, Diagnostics, Run Sanitization, Address Disinfec Sanitizer ion...

src/game/boe.newgraph.cpp
fosnola: I tried to convert the damage_pc code to show the maximum damage protection:
```
		for (i = 0; i < 24; i++)
			if ((adven[which_pc].items[i].variety != 0) && (adven[which_pc].equip[i] == TRUE)) {
				if ((adven[which_pc].items[i].variety >= 12) && (adven[which_pc].items[i].variety <= 17)) {
						r1 = get_ran(1,1,adven[which_pc].items[i].item_level);
						how_much -= r1;
						
						// bonus for magical items
						if (adven[which_pc].items[i].bonus > 0) {
							r1 = get_ran(1,1,adven[which_pc].items[i].bonus);
							how_much -= r1;
							how_much -= adven[which_pc].items[i].bonus / 2;
							}
						if (adven[which_pc].items[i].bonus < 0) {
							how_much = how_much - adven[which_pc].items[i].bonus;
							}
						r1 = get_ran(1,0,100);
						if (r1 < hit_chance[adven[which_pc].skills[8]] - 20)
							how_much -= 1;
					}
				if (adven[which_pc].items[i].protection > 0) {
						r1 = get_ran(1,1,adven[which_pc].items[i].protection);
						how_much -= r1;
					}
				if (adven[which_pc].items[i].protection < 0) {
						r1 = get_ran(1,1,-1 * adven[which_pc].items[i].protection);
						how_much += r1;
					}
				}	

```
but yes, I make some mistakes,  

- concerning item_level : if (item_level<=0) we add a number in [item_level:1] else [1, item_level]
- concerning the bonus : if (bonus<0) we add bonus else [1+bonus/2, 3*bonus/2]
- concerning the protection : if (protection<0) we add [protection,-1] else [1,protection]

CelticMinstrel:
Yes, I'm aware, thank you. I actually did that a little while ago to fix a crash in the vector2d unit tests.

src/scenario/scenario.cpp
CelticMinstrel: I'm guessing the other patch in question is this one?

3cc550b4d9b7aa9d7909bbb25022907df91d9810

src/game/boe.newgraph.cpp
CelticMinstrel: I think this should be correct, right?

68ef066dca17823ddf9b99f2de9c6e0902b0d6bf

src/game/boe.fileio.cpp
CelticMinstrel: > Yes, another (more elegant) solution would be to update univ.scenario.outdoors.maps before storing a game. So when we reload a game, we would not have to rebuild this map using univ.scenario.outdoors.maps (which it is almost ok) and out_e (which contains the current portion)

This does sound better and looks less strange. That would mean calling … `save_outdoor_maps()` just before saving the game?

src/game/boe.graphics.cpp
```diff
@@ -860,6 +860,13 @@ void draw_terrain(short	mode) {
 			
 			if(is_town() || is_combat())
 				draw_items(where_draw);
+			
+			int max_dim = is_town() ? int(univ.town->max_dim) : 96;
+			if (where_draw.x<0 || where_draw.x>=max_dim || where_draw.y<0 || where_draw.y>=max_dim) {
+				place_road(q,r,where_draw,false);
```
CelticMinstrel, line 857: ?

Why are you placing a road when it's out-of-bounds?

src/game/boe.graphutil.cpp
```diff
@@ -319,6 +319,11 @@ void draw_outd_boats(location center) {
 		if(boat.which_town != 200) continue;
 		if(univ.party.in_boat >= 0 && univ.party.boats[univ.party.in_boat] == boat) continue;
 		location loc = local_to_global(boat.loc);
+		if (is_out()) {
```
CelticMinstrel, line 229: This seems incorrect for two reasons.

1. Redundant `is_out()` check. This function is `draw_outd_boats`, which is called to draw boats if you're outdoors. It checks `is_out()` before calling it.
2. Is this really the only place this logic is needed? I wonder if perhaps it belongs in `local_to_global()`.

If I understand correctly, this is accounting for a case where the party and the boat/horse are in adjacent sectors (presumably both near the edge of the sector). But, outdoor wandering encounters probably need the same logic, don't they? And possibly other things I haven't thought of…

src/universe/population.cpp
```diff
@@ -15,56 +15,75 @@
...
 #include "oldstructs.hpp"
 
+cCreature cPopulation::dummyDude;
+
 void cPopulation::import_legacy(legacy::creature_list_type old){
 	dudes.resize(60);
-	for(int i = 0; i < 60; i++)
-		dudes[i].import_legacy(old.dudes[i]);
+	for(int i = 0; i < 60; i++) {
+		dudes[i]=std::make_shared<cCreature>();
```
CelticMinstrel, line 23: Hmm, I'm not too keen on storing every cCreature as a shared pointer. What are the problems that can occur when they're killed without this change?

fosnola: In many cases, the game does not check correctly if a creature is killed. So after killing a creature (and deleting its memory), it still changes some of its members. This is why many games/saves have become corrupted :-~

**Note** In practice, using a share_ptr seems to be the easiest way to avoid this problem because when you take care of a creature, many external things can happen (which can kill it) before you are done taking care of it.

src/game/boe.graphics.cpp
fosnola: I guess I have already protected place_road to avoid the problem when it is called when it is out of bounds. But yes, it would be better to decide what to do in this case and not call another function.

src/game/boe.fileio.cpp
fosnola: It is probably best to create a function (update_map or something like that) to execute
```
	location corner = univ.party.outdoor_corner;
	for(short i = 0; i < 48; i++)
		for(short j = 0; j < 48; j++) {
			univ.scenario.outdoors[corner.x][corner.y]->maps[j][i] = univ.out.out_e[i][j];
			if(corner.x + 1 < univ.scenario.outdoors.width())
				univ.scenario.outdoors[corner.x + 1][corner.y]->maps[j][i] = univ.out.out_e[i + 48][j];
			if(corner.y + 1 < univ.scenario.outdoors.height())
				univ.scenario.outdoors[corner.x][corner.y + 1]->maps[j][i] = univ.out.out_e[i][j + 48];
			if(corner.y + 1 < univ.scenario.outdoors.height() && corner.x + 1 < univ.scenario.outdoors.width())
				univ.scenario.outdoors[corner.x + 1][corner.y + 1]->maps[j][i] = univ.out.out_e[i + 48][j + 48];
		}
```
Then we can call this function before saving ( and we can modify the first part of save_outdoor_map to call this new function)

src/scenario/scenario.cpp
fosnola: Yes.

src/game/boe.graphutil.cpp
fosnola: Yes, you can suppress the redundant is_out().

**Note** I'm not a fan of these local_to_global/global_to_local methods either, they generate a lot of errors. Then I just tried to fix the code when I detected a problem, without really thinking if there was another solution less problematic.

src/game/boe.newgraph.cpp
fosnola: > Okay, so is your `item_level <= 0` check based on the fact that `get_ran` returns `times * min` if `max` is less than `min`? (That would also happen for `item_level == 0` though.)

Yes.

> 
> If 'bonus' and 'protection' are negative but 'item_level' is still positive, you get the following formula:
> 
> [1, item_level] - bonus - [1, -protection]
> 

This is [1, item_level] + bonus - [1, -protection]
( ie. we add -bonus damage, so the protection is negative : bonus )

> I think this should be correct, right?
> 
> [68ef066](https://github.com/calref/cboe/commit/68ef066dca17823ddf9b99f2de9c6e0902b0d6bf)

This seems to be correct. I don't know if there is enough room to display the result; if so, it might be better to display "minprotect - maxprotect" or "minprotect—maxproctect" to avoid displaying -3--2.

src/universe/population.cpp
CelticMinstrel: So basically you're saying there out-of-bounds writes on the `dudes` array. Okay.

src/game/boe.fileio.cpp
CelticMinstrel: Huh? But that's the entire content of `save_outdoor_maps()`?

src/universe/population.cpp
fosnola: Yes.

But there is also the case (less crash-generating) where some creatures are summoned after the death of a creature, i.e. if we continue to modify the members of the creature that has just died, we will modify some characteristics of a new summoned creature.

src/game/boe.fileio.cpp
fosnola: Oops, yes, this is the code of save_outdoor_maps ( so you can call save_outdoor_maps ).

**Note:** my code is different because I added the ability to also create cities that keep the terrain changes. For example, if you destroy a building, you can always return to the city and the building is still destroyed...

src/universe/party.cpp
```diff
@@ -792,9 +792,12 @@ void cParty::writeTo(std::ostream& file) const {
 		}
 	file << '\f';
 	file << '\f';
-	for(unsigned int i = 0; i < party_event_timers.size(); i++)
+	for(unsigned int i = 0; i < party_event_timers.size(); i++) {
+		if (party_event_timers[i].node<=-200 || (party_event_timers[i].node>=-9 && party_event_timers[i].node<0))
```
CelticMinstrel, line 1041: Can you explain this condition? Why are both not simply `<= 0`? Is there a reason `time` is not checked?

src/game/boe.fileio.cpp
CelticMinstrel: Right, so even after this pull request is finished, our two branches won't even be close to synced…

It's definitely good to deal with the more obvious fixes like this first before deciding what to do with actual new features you've implemented, though.

src/universe/population.cpp
CelticMinstrel: Hmm… leaving stale monsters in the array indefinitely is certainly a possibility, though that creates a chance that the `dudes` array grows without bound…

fosnola: To be compatible with older compilers, the simplest solution might be to use boost::shared_ptr ( which is the ancestor of std::shared_ptr ). 

Because using std::vector<cCreature> dudes and passing a reference to dudes[2] to a method is really dangerous: if the dudes vector grows (because a monster is summoned), the reference points anywhere. 

At the very least, replacing std::vector<cCreature> dudes with a std::array<cCreature, XXX> is maybe a bit healthy (even if you have to choose XXX and add protections in the code).

src/universe/party.cpp
fosnola: Yes, just check that party_event_timers.node<0 : I guess I was trying to remove the export of unnecessary timers, and wasn't sure if magic_ptrs and pointers were used when party_event_timers.node<0 :-~

**Note**: Of course, you can also check party_event_timers.node.time

src/universe/population.cpp
CelticMinstrel: > To be compatible with older compilers, the simplest solution might be to use boost::shared_ptr ( which is the ancestor of std::shared_ptr ).

I'm choosing not to be compatible with any compiler that doesn't support (at least most of) C++11. Therefore, no need for `boost::shared_ptr`. I'm already using `std::shared_ptr` in other places.

My objection here isn't to `std::shared_ptr` but to the idea of using a shared pointer (or even a unique pointer) at all.

> Because using _std::vector< cCreature >_ dudes and passing a reference to dudes[2] to a method is really dangerous: if the dudes vector grows (because a monster is summoned), the reference points anywhere.

Ah, do things keep references (or pointers) to elements of the creature array? If so, maybe we should play it safe and use `std::deque` instead… but that said, as long as the references don't outlive a function, it should still be fine with `std::vector`…

> At the very least, replacing std::vector dudes with a std::array<cCreature, XXX> is maybe a bit healthy (even if you have to choose XXX and add protections in the code).

This is an option as a last resort, but it puts us back in the old state where creatures have a hard upper limit. I could still choose a _much higher_ upper limit than the old game had, but…

src/universe/party.cpp
CelticMinstrel: Looks like pointers are not used in this context, so I'll just check `< 0` then. Node 0 should also be valid, after all.

src/universe/population.cpp
fosnola: For instance, if you look at the code:
```
void do_monster_turn() {
	cCreature *cur_monst;
....	
	for(short i = 0; i < num_monst; i++) {  // Give monsters ap's, check activity
		
		cur_monst = &univ.town.monst[i];
....
						acted_yet = monst_cast_mage(cur_monst,target);
....

```
this is problematic, because cur_monst keeps a pointer to univ.town.monst[i] and this pointer does not changed when univ.town.monst is resized (e.g., when a new monster is summoned) :-~

src/game/boe.actions.cpp
```diff
@@ -2976,10 +2977,10 @@ bool town_move_party(location destination,short forced) {
 		forced = true;
 	if(univ.debug_mode && univ.ghost_mode)
 		forced = keep_going = true;
-	
-	ter = univ.town->terrain(destination.x,destination.y);
-	
+
 	if(keep_going) {
+		// ASAN destination is only correct if keep_going=true
+		ter = univ.town->terrain(destination.x,destination.y);
```
CelticMinstrel, line 3061: Though it's correct to move this here, how can it solve an ASAN problem? There's nothing here or in `check_special_terrain` that verifies the location is actually on the map.

fosnola: Yes, it would be better to check that the destination is valid at the beginning of this function.

( I guess when asan detected a problem here, keep_going was false so I made this change but it is not a good change )

src/game/boe.items.cpp
```diff
@@ -313,17 +313,24 @@ void set_town_attitude(short lo,short hi,eAttitude att) {
...
-		lo = 0;
-	if(lo < 0)
-		lo = num_monst + lo;
-	if(hi <= -num_monst)
-		hi = 0;
-	if(hi < 0)
-		hi = num_monst + hi;
-	if(hi < lo)
-		std::swap(lo, hi);
+	if (lo==-1 && hi==-1) {
```
CelticMinstrel, line 342: Why does this mean all monsters? Is there something that uses this?

src/game/boe.specials.cpp
```diff
@@ -1917,7 +1917,11 @@ void queue_special(eSpecCtx mode, eSpecCtxType which_type, spec_num_t spec, loca
 	queued_special.type = which_type;
 	queued_special.mode = mode;
 	queued_special.trigger_time = univ.party.age;
-	special_queue.push(queued_special);
+	// FIXME: I forced calling the leave special just to avoid calling them outside, ie. with town_num=200
```
CelticMinstrel, line 1937: You call this a hack and leave a FIXME. Did you find a better solution later on?

src/game/boe.graphics.cpp
```diff
@@ -738,7 +738,7 @@ void draw_terrain(short	mode) {
 			where_draw.x += i - 6;
 			where_draw.y += j - 6;
 			if(!(is_out()))
-				light_area[i][j] = (is_town()) ? pt_in_light(view_loc,where_draw) : combat_pt_in_light(where_draw);
+				light_area[i][j] = (is_town()) ? (can_see_light(view_loc,where_draw,sight_obscurity)<=4) : combat_pt_in_light(where_draw);
```
CelticMinstrel: Hmm. Which of these is legacy behaviour in Blades of Exile? What about in Exile III? And what does this change actually look like?

Seeing explored regions (terrain only) through walls is expected behaviour, at least in towns with an inherent light level. I'm not sure if that applies to dark towns.

src/game/boe.specials.cpp
```diff
@@ -2812,7 +2812,13 @@ void affect_spec(const runtime_state& ctx) {
 				for(short i = 0; i < 6; i++)
 					if(pc_num == 6 || pc_num == i) {
 						if(spec.ex1b == 0) {
-							if(spec.ex1a == 3 && is_combat() && which_combat_type == 0 && univ.party[i].main_status == eMainStatus::FLED)
+							if(spec.ex1a == 0 && univ.party[i].main_status == eMainStatus::DEAD)
```
CelticMinstrel, line 2830: So, at first glance this seems to be unnecessary. The node already has a check to restore PCs that were dead, dust, or stoned – it's the final `else` in this chain. This change means you can only remove one of the three at a time, and that final fallback will never be hit unless `ex1a` is -1 or something, right?

When you wrote this, was it because raising dead with this node wasn't working at all? Because, if that's the case, that's a different fix from this.

On the other hand, if you've done this because you want a way to unstone PCs without also resurrecting dead PCs, then it seems like a reasonable thing to support.

src/game/boe.fileio.cpp
```diff
@@ -59,6 +60,7 @@ void finish_load_party(){
 	bool town_restore = univ.party.town_num < 200;
 	bool in_scen = univ.party.scen_name.length() > 0;
 	
+	end_scenario = false;
```
CelticMinstrel: I elected to put this in `handle_victory()` instead, but can you think of any reason that would be insufficient?

src/universe/pc.cpp
```diff
@@ -299,7 +299,6 @@ void cPlayer::acid(int how_much) {
 
 void cPlayer::restore_sp(int amt) {
 	if(!is_alive()) return;
-	if(cur_sp >= max_sp) return;
```
CelticMinstrel, line 302: I think I already mentioned that this is incorrect and the proper fix is to call `drain_sp` from the special node.

I imagine the reason that wasn't already done is because `drain_sp` implements resistance, and the goal was that the special node can't be resisted, so I also added a flag to disable resistance.

See d7dcf2464426496cb64d34faa77dccabad251ed7 for details.

src/fileio/fileio_party.cpp
```diff
@@ -264,7 +264,10 @@ bool load_party_v1(fs::path file_to_load, cUniverse& real_univ, bool town_restor
 					}
 				}
 			}
-		else univ.party.town_num = 200;
+		else {
+			univ.party.town_num = 200;
+			univ.town.difficulty = 0;
```
CelticMinstrel, line 251: I decided to delete this. It looked like it was just causing confusion, and we should be using `cTown::difficulty` (`univ.town->difficulty`) instead.

src/game/boe.town.cpp
```diff
@@ -1538,7 +1538,7 @@ void draw_map(bool need_refresh) {
 		}
 	}
 	
-	mini_map.setActive(false);
+	mini_map.setActive();
```
CelticMinstrel: Does this do something? I can't see any difference with or without this change, but maybe I'm missing something. It does seem like the change is correct, though…

src/dialogxml/dialogs/dialog.cpp
```diff
@@ -508,16 +508,31 @@ void cDialog::run(std::function<void(cDialog&)> onopen){
...
+		if (ui_scale < 0.1) ui_scale = 1.0;
+	}
+
+	// check that the dialog can be displayed on the screen
+	sf::VideoMode desktop = sf::VideoMode::getDesktopMode();
+	if (desktop.width< ui_scale*winRect.width())
+		ui_scale=float(desktop.width)/winRect.width();
+	// we need keep enough space for the mac menu and the dialog title
+	// and add more space on a retina mac
+	// checkme: we need probably to also add getMenuHeight at least on Linux
```
CelticMinstrel, line 526: I'd like to accept this change, but I have three requests:

* Unless you're certain it _creates_ an issue, I'd like to keep the hack that sleeps for 100ms and creates and destroys a 1x1 window. If it's not causing any problems and _might_ be fixing an issue for someone, it shouldn't just be deleted.
* There's no need to call getMenuHeight because only the main window has menus. (Edit menu shortcuts like copy and paste work independently of the actual menu.) Therefore, you can delete the checkme comment.
* In e47b7e6a223a4fbc666e04138f9b93aefe2b1002 I allowed differing the UI scale in the game and in the editors, so that will need to be handled somehow. Passing UI scale as a parameter in the constructor is one possibility, though that carries the annoying task of updating every single dialog construction in the entire source to pass it. Another possibility would be a `get_ui_scale()` function that's implemented differently in each program.

I'd also like to better understand what you're doing here to limit the size to something that fits on the screen. You're just scaling it by less than the expected 2x in that case, aren't you? I wonder if this is a good approach… doesn't 1.5x look a lot worse than 2x?

pkg/mac/BoE-Info.plist
```diff
@@ -239,5 +239,7 @@
 			</dict>
 		</dict>
 	</array>
+	<key>NSHighResolutionCapable</key>
```
CelticMinstrel, line 242: I understand that ultimately this is a good thing to do at some point, but what effect does it have? On my computer I can't see any difference. I assume it does make a difference on your computer though.

src/game/boe.actions.cpp
```diff
@@ -2588,7 +2588,7 @@ void handle_death() {
 			fs::path file_to_load = nav_get_party();
 			if(!file_to_load.empty()) load_party(file_to_load, univ);
 			if(univ.party.is_alive()) {
-				if(overall_mode != MODE_STARTUP)
+				if(overall_mode == MODE_STARTUP)
```
CelticMinstrel: This doesn't look right. `post_load()` shows various scrollbars to indicate that you're in the game, so it shouldn't be called if mode is `MODE_STARTUP`.

src/fileio/resmgr/res_image.cpp
```diff
@@ -8,12 +8,20 @@
 
 #include "res_image.hpp"
 
-class ImageLoader : public ResMgr::cLoader<sf::Texture> {
+#include "gfx/texture.hpp"
+
+class TextureLoader : public ResMgr::cLoader<Texture> {
```
CelticMinstrel, line 13: I don't want to accept this patch.

I'm not sure what it does, and a different patch that does a similar thing may be acceptable as long as I can understand its purpose.

However, this patch renames a whole bunch of things that didn't really need to be renamed at all. You could have easily rewritten the internals of this loader without renaming it, its instance, or the `using` typedef for the resource pool's return type.

If you want to resubmit this later after renaming all those things (and anything else you renamed) back to their original names, and you can explain what it does and why, then I'll consider it. But I won't accept it its current form.

Also, if you do end up doing that, please submit it in a pull request all on its own.

src/scenario/terrain.cpp
```diff
@@ -111,12 +111,12 @@ void cTerrain::import_legacy(legacy::terrain_type_type& old){
...
 	is_archetype = archetypes.count(picture) && (old.special == 0) == (old.picture != 404);
 	blockage = (eTerObstruct) old.blockage;
-	if(picture < 260){
+	if(picture>=0 && picture < 260){
 		combat_arena = arenas[picture];
 		ground_type = ground[picture];
 		trim_type = (eTrimType) trims[picture];
 		trim_ter = trim_ters[picture];
-	}else if(picture >= 400 && picture < 1000){
+	}else if(picture >= 400 && picture < 140+274){
```
CelticMinstrel, line 119: Why such a strange number as `140+274`? What's wrong with `1000`?

src/fileio/fileio_scen.cpp
```diff
@@ -155,6 +155,33 @@ template<typename Container> static void port_shop_spec_node(cSpecial& spec, std
 	spec.ex2a = spec.ex2b = -1;
 }
 
+static void port_special_text_response(cSpecial &special, cScenario& scenario)
```
CelticMinstrel, line 157: Can you explain what this is doing and why?

src/fileio/fileio_party.cpp
```diff
@@ -209,11 +210,11 @@ bool load_party_v1(fs::path file_to_load, cUniverse& real_univ, bool town_restor
 			fin.read((char*)&o_maps, len);
 		}
 		
-		// LOAD SFX & MISC_I
+		// SKIP SFX & MISC_I
```
CelticMinstrel, line 191: Why skip them?

src/game/boe.dlgutil.cpp
```diff
@@ -194,11 +194,15 @@ void end_shop_mode() {
...
+		          as in start_shop_mode, store_limited_stock is checked before the
+				  item is loaded from the magic store magic_store_items
+		 */
 		for(int i = 0; i < active_shop.size(); i++) {
 			cShopItem item = active_shop.getItem(i);
-			if(item.type == eShopItemType::TREASURE || item.type == eShopItemType::CLASS || item.type == eShopItemType::OPT_ITEM)
-				this_shop[i] = item.item;
+			if(item.quantity>0 && item.type == eShopItemType::EMPTY &&
+			   this_shop.find(i)!=this_shop.end())
+				this_shop.find(i)->second.variety = eItemType::NO_ITEM;
```
CelticMinstrel, line 212: Can you explain the reasoning behind this change? You seem to be completely changing the logic.

src/game/boe.newgraph.cpp
```diff
@@ -399,6 +399,8 @@ void do_missile_anim(short num_steps,location missile_origin,short sound_num) {
 	// Now, at last, launch missile
 	for(short t = 0; t < num_steps; t++) {
 		draw_terrain();
+		// prevent drawing to happen outside the terrain windows
+		clip_rect(mainPtr, terrain_screen_rect);
```
CelticMinstrel, line 404: Unfortunately this doesn't compile after cherry-picking it… I assume it depends on some other commit that I didn't take…

src/game/boe.actions.cpp
```diff
@@ -1590,12 +1590,39 @@ bool handle_keystroke(const sf::Event& event){
...
 	Key shop_chars[8] = {kb::A,kb::B,kb::C,kb::D,kb::E,kb::F,kb::G,kb::H};
 	
-	if(map_visible && event.key.code == kb::Escape
-	   && (overall_mode != MODE_TALKING) && (overall_mode != MODE_SHOPPING)) {
-		mini_map.setVisible(false);
-		map_visible = false;
-		mainPtr.setActive();
-		return false;
+	if(event.key.code == kb::Escape) {
+		bool abort=true;
```
CelticMinstrel, line 1597: I think this is broadly a good idea, but I'm wondering if you might have missed a detail somewhere. Most of these modes have another way to cancel them, don't they? Usually by pressing the same key again? Does cancelling with Escape have the exact same outcome as cancelling normally?

I was trying to determine this for myself and found something that suggested it might not be the case for at least `MODE_TOWN_TARGET`, but I didn't look much further beyond that.

Ultimately, this is a good idea but probably not a good time to do it. We need to refactor the input so that the action and the input that triggers it aren't so tightly coupled. Then implementing this would probably be much easier to do correctly.

src/pcedit/pc.editors.cpp
```diff
@@ -306,9 +306,10 @@ static bool pick_alchemy_led(cDialog& me, std::string item_hit, bool, bool allow
 		cLed& led = dynamic_cast<cLed&>(me.getControl(item_hit));
 		led.setState(univ.party.alchemy[hit] ? led_red : led_off);
 	}
-	// TODO: change message when hit==0
-	//       use a variable to store the beginning of the potion details
-	me["info"].setText(get_str("alchemy", hit+8));
+	if (hit==0)
+		me["info"].setText("Weak Curing\n  Min. Skill - 1\nNeed holly. Cures poison somewhat.");
```
CelticMinstrel, line 305: This is definitely wrong. The string shouldn't be hard-coded into the source code like this.

I think the correct answer is to `get_str` 8 and delete everything up to the first `|`. An alternate answer would be to split that string in two and add 9 instead of 8.

src/pcedit/pc.editors.cpp
```diff
@@ -46,88 +46,83 @@ extern short d_rect_index[80];
...
+	me["who"].setText(univ.party[pc_num].name.c_str());
 }
 
-static bool display_pc_event_filter(cDialog& me, std::string item_hit, const short trait_mode) {
-	short pc_num;
-	
-	pc_num = which_pc_displayed;
-	if(item_hit == "done") {
+static bool display_pc_event_filter(cDialog& me, std::string item_hit, short &pc_num, const short mode, bool edit) {
+	if (edit) {
```
CelticMinstrel, line 70: No, I'm pretty sure this is wrong. Doesn't this save every time the user clicks an LED?

CelticMinstrel:
Okay, I got roughly as far as cc8b9be4fee667e63376a5f1323d61ee3574c77c, at which point I start running into a lot of conflicts. It's likely because I skipped the textures commit.

It would be nice if you could rebase on the latest master again and resolve the conflicts. Many of the commits would probably be dropped automatically, and there are also several that have been addressed in other ways which should also be dropped. I commented on some of them in this thread, but I may not have commented on all of them. In particular, if it's an asan commit, it's quite likely I already did something else to address the issue.

I'm going to close this now. We can continue to discuss the specific commits that have open threads on them in here, and once you've rebased your work you can open a new pull request.

src/game/boe.items.cpp
fosnola: I probably tried to correct a problem with the old special 170 which normally called :
```
		case 170:
			make_town_hostile();
			break;
```
currently we have
```
		case 170: type = eSpecType::MAKE_TOWN_HOSTILE; break;
			// 171-173 were change/swap/transform terrain
```
and
```
		case eSpecType::MAKE_TOWN_HOSTILE:
			if(spec.ex2a < 0 || spec.ex2a > 3){
				showError("Invalid attitude (0-Friendly Docile, 1-Hostile A, 2-Friendly Will Fight, 3-Hostile B).");
				break;
			}
			set_town_attitude(spec.ex1a,spec.ex1b,eAttitude(spec.ex2a));
			break;
```

So if the code does no change the values of ex1a and ex1b, we call set_town_attitude(-1,-1, eAttitude(spec.ex2a)); which kills almost nobody. But yes, it would be better to change:
```
		case 170: type = eSpecType::MAKE_TOWN_HOSTILE; ex1a=0; ex1b=-1; break;
```

src/game/boe.specials.cpp
fosnola: I keep this modification, it works. But I still think it's a hack (not an example of good programming).

src/game/boe.graphics.cpp
fosnola: I don't know why, but I keep seeing regions explored in cities with an inherent level of light.

Also note that after in the code we have :
```
			if(!(is_out()) && ((where_draw.x < 0) || (where_draw.x > univ.town->max_dim - 1)
								|| (where_draw.y < 0) || (where_draw.y > univ.town->max_dim - 1))) {
...
				if(can_see_light(view_loc,where_draw,sight_obscurity) < 5)
					can_draw = 1;
				else can_draw = 0;

```
so the use of can_see_light seems more natural (and when I made this change, I was probably correcting a problem in dark cities)

src/game/boe.specials.cpp
fosnola: In specials-text-affect.txt, you have:
```
Kill/Raise Dead
Unused
Unused
First part of message
Second part of message
Unused
Unused
Unused
0 - dead, 1 - dust,  2- stoned
0 - raise dead, 1 - hurt
Unused
Unused
Unused
Unused
```
which seems to suggest that if ex1a=0, we only want to raise the dead, if ex1a=1 the dust, ... So we need to add these cases. 

**Note** when converting a legacy game, we need to set ex1a to -1, so it will resurrect all people.

src/game/boe.fileio.cpp
fosnola: Yes, it is probably better.

src/universe/pc.cpp
fosnola: Ok.

src/fileio/fileio_party.cpp
fosnola: Yes, this seems better.

src/game/boe.town.cpp
fosnola: My solution to avoid flickering was to call setActive(false); do change; setActive() and there I make a mistake: I write setActive(false); do change; setActive(false), so I want to correct this mistake. Here, I suppose that the call after mini_map.display() is enough.

src/dialogxml/dialogs/dialog.cpp
fosnola: > I'd like to accept this change, but I have ~three~two requests:
> 
> * Unless you're certain it _creates_ an issue, I'd like to keep the hack that sleeps for 100ms and creates and destroys a 1x1 window. If it's not causing any problems and _might_ be fixing an issue for someone, it shouldn't just be deleted.

I know that before I changed this code completely, the hack was necessary on my mac and with the new code, it is not necessary anymore. So I guess I solved this problem but we can keep the hack in a dead code (with #if 0 #endif), so we can reactivate it if necessary.

> * There's no need to call getMenuHeight because only the main window has menus. (Edit menu shortcuts like copy and paste work independently of the actual menu.) Therefore, you can delete the checkme comment.

Ok

> * ~In [e47b7e6](https://github.com/calref/cboe/commit/e47b7e6a223a4fbc666e04138f9b93aefe2b1002) I allowed differing the UI scale in the game and in the editors, so that will need to be handled somehow. Passing UI scale as a parameter in the constructor is one possibility, though that carries the annoying task of updating every single dialog construction in the entire source to pass it. Another possibility would be a `get_ui_scale()` function that's implemented differently in each program.~
>
 
I have to investigate. Actually, in my code, it's also possible, at least on Mac, ie. each executable uses uses a different preference file. We'd have to check if this is true on Windows and on Linux.

> I'd also like to better understand what you're doing here to limit the size to something that fits on the screen. You're just scaling it by less than the expected 2x in that case, aren't you? I wonder if this is a good approach… doesn't 1.5x look a lot worse than 2x?

Yes, the appearance is a bit worse, but having to reduce the size of a dialog box almost never happens. So in practice, using a size closed to the desired size seems ok....

pkg/mac/BoE-Info.plist
fosnola: When I added this flag, I noticed some small changes, https://developer.apple.com/library/archive/documentation/GraphicsAnimation/Conceptual/HighResolutionOSX/Explained/Explained.html . But it is not impossible that the new versions of the system have made these changes minimal.

src/game/boe.actions.cpp
fosnola: The current code sets overall_mode to MODE_STARTUP:
```
void handle_death() {
	std::string choice;
	
	overall_mode = MODE_STARTUP;
	
	while(true) {
..
			fs::path file_to_load = nav_get_party();
			if(!file_to_load.empty()) load_party(file_to_load, univ);
			if(univ.party.is_alive()) {
				if(overall_mode != MODE_STARTUP)
```
and if I don't make an error, load_party never changes overall_mode. So when a party is loaded (and alive), we always have overall_mode == MODE_STARTUP but we still have to finish loading the party.

Ie. Basically, this patch should make the code work, but this code should probably be completely rewritten.

src/scenario/terrain.cpp
fosnola: Because we have only 274 arenas:
`	static const short arenas[274] = {`

src/fileio/resmgr/res_image.cpp
fosnola: For me (and SFML), we are mainly reading textures here. And using ImageLoader with ResMgr::graphics doesn't seem consistent.

But I'll try to rewrite this code to keep the old name (when I find time to do it).

src/fileio/fileio_scen.cpp
fosnola: In a legacy scenario, all the strings are stored in one block. But load_scenario_v1 split them in different variables
`
	for(short i = 0; i < 270; i++) {
		len = (long) (temp_scenario.scen_str_len[i]);
		fread(temp_str, len, 1, file_id);
		temp_str[len] = 0;
		if(i == 0) scenario.scen_name = temp_str;
		else if(i == 1 || i == 2)
			scenario.who_wrote[i-1] = temp_str;
		else if(i == 3)
			scenario.contact_info[1] = temp_str;
                ...
`
So, after uploading the scenario, we need to check that a string referenced by a special node was not referencing an unusual string (and if so, correct the problem).

src/pcedit/pc.editors.cpp
fosnola: Yes, it will be better to simply use strings 8...

src/game/boe.actions.cpp
fosnola: Yes, you can always cancel an action by pressing the same key again. But in practice, when I want to undo an action, I find it easier to always press the Escape key than to have to search for the key I need to press to undo that action.

fosnola:
> It would be nice if you could rebase on the latest master again and resolve the conflicts. Many of the commits would probably be dropped automatically, and there are also several that have been addressed in other ways which should also be dropped. I commented on some of them in this thread, but I may not have commented on all of them. In particular, if it's an asan commit, it's quite likely I already did something else to address the issue.
> 

I will try to rebase it next week ( but it may take some times to do that ) and I will try to answer tomorrow the two points I left unanswered...

src/game/boe.specials.cpp
CelticMinstrel: That _does_ sound logical, but according to [the documentation](http://openboe.com/docs/editor/appendix/Specials.html?jumpto=64), Extra 1a is ignored when Extra 1b is non-zero, though I think it's a mistake and it meant to say when Extra 1b is _zero_.

As you say, if we make this change, Extra 1a would need to be set to -1 or similar in a legacy game so that it hits the fallback that cures dead, dust, _and_ stone. Also, the HTML documentation would need to be updated.

src/universe/population.cpp
CelticMinstrel: Okay. I think this can be solved by using `std::deque` instead of `std::vector`, as long as it's always resized only by adding a new one on the end.

src/game/boe.town.cpp
CelticMinstrel: I believe calling `display()` may implicitly set it active first if it isn't already, so it's probably just redundant.

src/dialogxml/dialogs/dialog.cpp
CelticMinstrel: > I have to investigate. Actually, in my code, it's also possible, at least on Mac, ie. each executable uses uses a different preference file. We'd have to check if this is true on Windows and on Linux.

Ah, yes, we'd have to make sure they also use a separate preference file on Windows and Linux. I reverted it because I noticed they already use a separate preference file on the Mac, but I'd have to check if that's also true on Windows and Linux.

> Yes, the appearance is a bit worse, but having to reduce the size of a dialog box almost never happens. So in practice, using a size closed to the desired size seems ok....

Have you seen cases where the dialog's size needs to be reduced when set to 2x? (I won't worry about when it's set to 4x.)

src/scenario/terrain.cpp
CelticMinstrel: Ahh, I see.

Hmm. Legacy BoE had only 14 animated terrain graphics, so it's not as though this would break anything. If something used a graphic number between 414 and 1000, then that would be an error on the scenario's part.

So I guess this is fine.

src/fileio/fileio_scen.cpp
CelticMinstrel: Is it actually possible for a special node to reference an unusual string? Have you confirmed that this works in  legacy BoE?

src/game/boe.actions.cpp
CelticMinstrel: If I'm not mistaken, `load_party` never changes the mode, right? Or am I missing something?

Maybe the correct fix is this?

```cpp
				finish_load_party();
				if(overall_mode != MODE_STARTUP)
					post_load();
```

I'm not sure if swapping the order those are called in works though.

pkg/mac/BoE-Info.plist
CelticMinstrel: Even after setting this flag, it appears the game is locked to "low resolution":

image

(That's from selecting the .app in the Finder and choosing Get Info.)

src/game/boe.graphics.cpp
CelticMinstrel: I checked the original code and it does use `pt_in_light` here, so we'd have to decide whether that's a bug in the original game (and thus needs fixing) or a quirk of the original game (so no fix required).

src/game/boe.items.cpp
CelticMinstrel: Okay, I went ahead and did that.

src/game/boe.actions.cpp
CelticMinstrel: I've added a bounds check both here and in `outd_move_party`.

src/universe/population.cpp
CelticMinstrel: I've changed it to a `std::deque`. Hopefully that'll be enough to fix any issues you saw, as references to elements of a deque remain valid when a new element is added to the end.

src/game/boe.graphutil.cpp
CelticMinstrel: I've applied this without the redundant `is_out()`. A glance at `draw_monsters()` suggests that it's probably handled there somehow too, but it's something to look out for.

src/dialogxml/dialogs/dialog.cpp
CelticMinstrel: I thought "View Changes" on this conversation took you to the commit I'd commented on, but apparently that's not true, so here it is: 5286e798408c66433949a5a5bf26c462e4a45403

src/fileio/resmgr/res_image.cpp
CelticMinstrel: For reference, this is the specific commit in case you're having trouble finding it: 848325b6eeafa33e20b074cafb352edb944a549b

src/fileio/fileio_party.cpp
CelticMinstrel: This is the commit: 441ebb358f9312be01cbbaa286167d5bc69adf7d

src/game/boe.dlgutil.cpp
CelticMinstrel: This is the commit: d92113d0869e99aac87660d859e354cdab0c2c2b

I'm ignoring all the porting stuff in that commit for now.

src/game/boe.newgraph.cpp
CelticMinstrel: This is the commit: 8c5938d12081f57f0d20f11ead9a5bcee2a40c38

src/game/boe.town.cpp
fosnola: Yes, I also imagine that this is the case.

src/dialogxml/dialogs/dialog.cpp
fosnola: On my macbook air (and on an iMac), I use 4x scaling. I think BoE preferences are scaled down a bit on my MacBook Air (maybe by a factor of 20%) but other than that, I rarely see scaled dialogs. So with 2x scaling, I'd say it should never happen.

pkg/mac/BoE-Info.plist
fosnola: Unsure, on OSX 12.6.1, I obtain for instance:
Capture d’écran 2023-01-23 à 09 52 09
which was created with
Capture d’écran 2023-01-23 à 10 04 12

src/game/boe.dlgutil.cpp
fosnola: There was a problem with the random store; the first time you start shopping, you see the list of items, but if you finish and start shopping again: there are no items.

Basically, modifying this_shop[i] is not valid, because this_shop[i].type will not be valid anymore (i.e. it will not be TREASURE, ...). Here I tried to implement:
`// If it was a random shop, we need to update the stored list of items so that bought items don't reappear
`

When the last item is purchased, takeOne sets item.type to eShopItemType::EMPTY ( and leaves item.quantity=1) => the new condition. And I assume that by setting the variety to NO_ITEM, the item will be hidden.

src/game/boe.actions.cpp
fosnola: Yes, it should work. Ie., post_load seems to mainly update the screen (and do some cleaning).

src/fileio/fileio_party.cpp
fosnola: Yes, I made a mistake by skipping SFX & MISC_I, i.e. I probably thought the old code called `load_town` with mode=0 (but it only calls it with mode=1 and mode =2) => I will undo that.

The other patches are basic cleanups and fix some errors: import of special_item, check that there are no duplicate creatures for the same town, add some forgotten flip_short, use out_e instead of expl (because out_e replaced expl), ...

src/dialogxml/dialogs/dialog.cpp
CelticMinstrel: So pick spell and training dialogs are not scaled down even at 4x?

fosnola: I'll add a breakpoint tomorrow to give an accurate answer. But if I remember what happened before I added this code to scale down the overly large dialogs, the spellcasting dialog was a bit too big (i.e., the ok button didn't show up completely), so it's probably also scaled down a bit. I don't remember if there was problem with the training dialogue.

fosnola: So I add a message when the dialogs are reduced: the preferences dialog is reduced ( uiScale=2.98), the casting and training dialogs are not reduced.
Capture d’écran 2023-01-24 à 18 25 29

Issue #244, opened 2020-01-27
Multiple Scenario Editor crashes


x-qq:
* Scenario -> Advanced -> "Import Outdoor Sector" and then canceling file selection dialog
* Scenario -> Advanced -> "Set Scenario Event Timers" and press "ok" without doing anything

NQNStudios:
I'm guessing that the first crash in this list is fixed now along with #301

NQNStudios:
I reproduced the second crash on Windows. It came with an error message:

![imagen](https://user-images.githubusercontent.com/2584745/212164626-274b4235-01a6-48f6-b762-454b21f6db5d.png)

NQNStudios:
The crash also happens if you *do* set a scenario event timer.

NQNStudios:
It's happening here: https://github.com/calref/cboe/blob/master/src/scenedit/scen.core.cpp#L3145

`i + 1` is making it try to access time11, so I think it's an off-by-1 error.

NQNStudios:
That was wrong, sorry. It's here: https://github.com/calref/cboe/blob/master/src/scenedit/scen.core.cpp#L3109

Looks like the UI allows editing 10 when there are 20 (there's a TODO at the location of my last comment indicating this) but the save function is trying to disconnect UI events from 20 controls.

NQNStudios:
I think a stack widget could be added to hold 2 pages of fields for 10 events each.

NQNStudios:
> I'm guessing that the first crash in this list is fixed now along with #301

On this point I was also wrong. Windows and Linux both crash with an error message when you cancel the file browser to import outdoor sections AND import towns.

NQNStudios:
This issue can be closed now.

PR #309, opened 2023-01-14
[WIP] Fix 244b

NQNStudios:
Still in progress--this PR will fix the second crash reported in #244 by adding an extra page of scenario event timers in the scenario editor so all 20 can be used.

The first 2 commits, cc18c56866fa0c90239789eb39c0ba4b931b0825 and 1c775da7d2fec3b711861bd4396e3a1181191cf4, fix bugs in cDialog that caused other issues which I didn't open tickets for:

1) text field controls inside stack widgets didn't become focused when clicked on. The tab key was the only way to focus them.
2) entering characters to a text field in a stack widget would cause a crash.

Those bugs could be observed by editing town details and trying to edit the 3 comment fields which are managed by a stack.

### DIFF
--- a/src/dialogxml/dialogs/dialog.cpp
+++ b/src/dialogxml/dialogs/dialog.cpp
@@ -412,8 +412,16 @@ void cDialog::loadFromFile(const DialogDefn& file){
 	currentFocus = "";
 	for(ctrlIter iter = controls.begin(); iter != controls.end(); iter++){
 		if(typeid(iter->second) == typeid(cTextField*)){
-			if(currentFocus.empty()) currentFocus = iter->first;
+			currentFocus = iter->first;
 			break;
+		}
+		if(iter->second->isContainer()){
+			cContainer* tmp = dynamic_cast<cContainer*>(iter->second);
+			tmp->forEach([this, iter](std::string key, cControl& child) {
+				if (typeid(&child) == typeid(cTextField*)) {
+					if (currentFocus.empty()) currentFocus = iter->first;
+				}
+			});
 		}
 	}
 }
@@ -485,7 +493,7 @@ bool cDialog::sendInput(cKey key) {
 	if(topWindow == nullptr) return false;
 	std::string field = topWindow->currentFocus;
 	if(field.empty()) return true;
-	dynamic_cast<cTextField*>(topWindow->controls[field])->handleInput(key);
+	dynamic_cast<cTextField&>(topWindow->getControl(field)).handleInput(key);
 	return true;
 }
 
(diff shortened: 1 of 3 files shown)


src/dialogxml/dialogs/dialog.cpp
```diff
@@ -771,13 +771,28 @@ bool cDialog::setFocus(cTextField* newFocus, bool force) {
...
+	for(auto iter = controls.begin(); iter != controls.end(); iter++) {
+		if(iter->second == newFocus) {
+			if (!force && !newFocus->triggerFocusHandler(*this, iter->first, false)) return false;
+			currentFocus = iter->first;
+			return true;
+		}
+		if(iter->second->isContainer()) {
+			cContainer* tmp = dynamic_cast<cContainer*>(iter->second);
+			std::string foundKey = "";
+			tmp->forEach([&](std::string key, cControl& child) {
```
CelticMinstrel, line 783: Generally I prefer to explicitly list what needs to be captured, rather than using `[&]` or `[=]` to capture everything.

CelticMinstrel:
Looks good so far.

NQNStudios:
038baddb02a99c0c11b71723fd7296c450d019f4 declares the captures on the closure you pointed out.

26b6d92ec32ca77d329681e492bd4f97f52726e9 fixes 2 more places where cDialog wasn't aware of text fields in containers.

https://github.com/calref/cboe/pull/309/commits/b2ce1f18719cc21a7ffef4afacd958c252371304 actually adds the stack and page buttons to allow editing the other 10 scenario event timers. I've tested on Windows to make sure it works and preserves the values across the 2 pages.

Fixes part 2 of #244.

src/dialogxml/dialogs/dialog.cpp
```diff
@@ -412,9 +412,17 @@ void cDialog::loadFromFile(const DialogDefn& file){
 	currentFocus = "";
 	for(ctrlIter iter = controls.begin(); iter != controls.end(); iter++){
 		if(typeid(iter->second) == typeid(cTextField*)){
-			if(currentFocus.empty()) currentFocus = iter->first;
+			currentFocus = iter->first;
 			break;
 		}
+		if(iter->second->isContainer()){
+			cContainer* tmp = dynamic_cast<cContainer*>(iter->second);
+			tmp->forEach([&](std::string key, cControl& child) {
```
CelticMinstrel, line 420: Same as before.

src/scenedit/scen.core.cpp
```diff
@@ -3139,17 +3142,24 @@ void edit_scenario_events() {
 	using namespace std::placeholders;
 	
 	cDialog evt_dlg(*ResMgr::dialogs.get("edit-scenario-events"));
+	cStack& stk = dynamic_cast<cStack&>(evt_dlg["list"]);
+	evt_dlg["prev"].attachClickHandler(std::bind([&] { return stk.setPage(0); }));
+	evt_dlg["next"].attachClickHandler(std::bind([&] { return stk.setPage(1); }));
```
CelticMinstrel, line 3147: There's no reason to use `std::bind` on a lambda like this. Just use the lambda. (Also, `[&stk]` please.)

src/dialogxml/dialogs/dialog.cpp
```diff
@@ -417,7 +417,7 @@ void cDialog::loadFromFile(const DialogDefn& file){
 		}
 		if(iter->second->isContainer()){
 			cContainer* tmp = dynamic_cast<cContainer*>(iter->second);
-			tmp->forEach([&](std::string key, cControl& child) {
+			tmp->forEach([&, iter](std::string key, cControl& child) {
```
CelticMinstrel, line 420: I think this should be `[this, iter]`?

Issue #324, opened 2023-01-25
Items are wiped when loading a save file.


NQNStudios:
On Linux, when I create a new party in-game, save, close the game, open it and load the party, they won't have any items.

The same goes for opening any party in the character editor.

CelticMinstrel:
This bug is quite mystifying. As far as I can tell, it goes wrong right here:

https://github.com/calref/cboe/blob/f083965e80ef7e6ba76ca8ef966d1b3396ac00af/src/fileio/fileio_party.cpp#L271

But how it can possibly go wrong there, I've yet to figure out.

NQNStudios:
I put a breakpoint on that line and the debugger didn't stop at it. Maybe that's not where it goes wrong?

NQNStudios:
Ah the breakpoint worked when I put in on the corresponding line in the v2 function.

NQNStudios:
More specifically, it goes wrong when the adventurers array is swapped, like we talked about on IRC: https://github.com/calref/cboe/blob/32f24b40d6da3d483f5d05f1212d0e1ae8223497/src/universe/party.cpp#L167

In my debugger when I open the party, I'm actually finding that a breakpoint on that line triggers TWICE. The first time, items are preserved. The second time, they get wiped. Are we expecting it to be called twice?

CelticMinstrel:
Fixed in #328.

PR #328, opened 2023-01-27
Fix swap() double-call

NQNStudios:
image

After finding that #324 was due to swap() getting called twice, I figured out why the double-call was happening. For some reason, making the argument to the cUniverse copy constructor a reference and not a double-reference averts the bug. The double-reference looked wrong to me anyway, so maybe this is a correct solution.

### DIFF
--- a/src/universe/party.cpp
+++ b/src/universe/party.cpp
@@ -164,7 +164,6 @@ void swap(cParty& lhs, cParty& rhs) {
 	swap(lhs.total_xp_gained, rhs.total_xp_gained);
 	swap(lhs.total_dam_taken, rhs.total_dam_taken);
 	swap(lhs.scen_name, rhs.scen_name);
-	swap(lhs.adven, rhs.adven);
 	swap(lhs.setup, rhs.setup);
 	swap(lhs.stored_items, rhs.stored_items);
 	swap(lhs.summons, rhs.summons);


NQNStudios:
My fix might break other places where the universe changes, because now although parties have their items when they load, entering VoDT or On a Ship to Algiers and taking a single step causes this error:

```
Unrecognized node type found: -256
        Note: This could indicate corruption in the scenario, but more likely is just a result of garbage data in unused nodes or in the memory structures they were read into. The unrecognized node type has been replaced with invalid type -1.Unrecognized node type found: -256
        Note: This could indicate corruption in the scenario, but more likely is just a result of garbage data in unused nodes or in the memory structures they were read into. The unrecognized node type has been replaced with invalid type -1.Unrecognized node type found: -256
        Note: This could indicate corruption in the scenario, but more likely is just a result of garbage data in unused nodes or in the memory structures they were read into. The unrecognized node type has been replaced with invalid type -1.Unrecognized node type found: 20480
        Note: This could indicate corruption in the scenario, but more likely is just a result of garbage data in unused nodes or in the memory structures they were read into. The unrecognized node type has been replaced with invalid type -1.Unrecognized node type found: 20992
        Note: This could indicate corruption in the scenario, but more likely is just a result of garbage data in unused nodes or in the memory structures they were read into. The unrecognized node type has been replaced with invalid type -1.Unrecognized node type found: 21248
        Note: This could indicate corruption in the scenario, but more likely is just a result of garbage data in unused nodes or in the memory structures they were read into. The unrecognized node type has been replaced with invalid type -1.Unrecognized node type found: -256
        Note: This could indicate corruption in the scenario, but more likely is just a result of garbage data in unused nodes or in the memory structures they were read into. The unrecognized node type has been replaced with invalid type -1.Unrecognized node type found: -256
        Note: This could indicate corruption in the scenario, but more likely is just a result of garbage data in unused nodes or in the memory structures they were read into. The unrecognized node type has been replaced with invalid type -1.Unrecognized node type found: -256
        Note: This could indicate corruption in the scenario, but more likely is just a result of garbage data in unused nodes or in the memory structures they were read into. The unrecognized node type has been replaced with invalid type -1.Unrecognized node type found: -256
        Note: This could indicate corruption in the scenario, but more likely is just a result of garbage data in unused nodes or in the memory structures they were read into. The unrecognized node type has been replaced with invalid type -1.Unrecognized node type found: -256
        Note: This could indicate corruption in the scenario, but more likely is just a result of garbage data in unused nodes or in the memory structures they were read into. The unrecognized node type has been replaced with invalid type -1.Unrecognized node type found: -256
        Note: This could indicate corruption in the scenario, but more likely is just a result of garbage data in unused nodes or in the memory structures they were read into. The unrecognized node type has been replaced with invalid type -1.Unrecognized node type found: -256
        Note: This could indicate corruption in the scenario, but more likely is just a result of garbage data in unused nodes or in the memory structures they were read into. The unrecognized node type has been replaced with invalid type -1.Unrecognized node type found: -256
        Note: This could indicate corruption in the scenario, but more likely is just a result of garbage data in unused nodes or in the memory structures they were read into. The unrecognized node type has been replaced with invalid type -1.Unrecognized node type found: -256
        Note: This could indicate corruption in the scenario, but more likely is just a result of garbage data in unused nodes or in the memory structures they were read into. The unrecognized node type has been replaced with invalid type -1.Unrecognized node type found: 23552
        Note: This could indicate corruption in the scenario, but more likely is just a result of garbage data in unused nodes or in the memory structures they were read into. The unrecognized node type has been replaced with invalid type -1.Loading scenario graphics... ("/home/nat/.config/openboe/blades/Scenarios/On a Ship to Algiers/Algiers.exs")
./test.sh: line 1: 342298 Segmentation fault      (core dumped) "./Blades of Exile"
```

NQNStudios:
Ah, my fix isn't the cause of that problem. It can be reproduced on master.

CelticMinstrel:
That problem is probably the error I mentioned earlier – when swapping the PCs it should swap the contents of the PCs, not the pointers.

CelticMinstrel:
So, it's great that we found something that works, but I am still a little dubious about this change. The constructor you changed is _supposed_ to be a move constructor, which takes an r-value reference (double `&&`). So while it's nice that this works, I think there must be something else wrong.

NQNStudios:
That makes sense. I think what the debugger revealed is that the
problematic line of code is calling both the move constructor AND
std::move, which seems wrong. Changing the constructor signature must have
caused it not to be called.

So I wonder if it would be more proper to make the load function’s universe
argument into a pointer and assign the pointer’s value to std:: move
instead. Maybe that will avoid calling the move constructor. That was the
first approach I took but it introduced other compiler errors so I put a
pin in it.

CelticMinstrel:
> I think what the debugger revealed is that the problematic line of code is calling both the move constructor AND std::move, which seems wrong.

????

`std::move` does not do anything. It just returns its argument unchanged. Its only purpose is to convert an l-value reference to an r-value reference, which will then trigger the compiler to invoke the move constructor instead of the copy constructor.

By changing the signature of the move constructor, it no longer qualifies as a move constructor, so the compiler is probably calling the copy constructor instead.

> Maybe that will avoid calling the move constructor.

I don't think avoiding the move constructor should be the goal here. Perhaps it's not correct for the move constructor to call swap in the first place. I'm not really sure though… it _sounds_ correct to me…

Note that when I replaced `real_univ = std::move(univ)` with `swap(real_univ, univ)`, I still got the bug.

NQNStudios:
Ok disregard what I said about std::move, I thought it actually moved the
data by calling swap. I don’t understand this stuff very well.

Swap() is definitely getting called twice, though…

NQNStudios:
I will share the actual callstacks of the two swap calls when I get a
chance

NQNStudios:
image
image

NQNStudios:
party.cpp
Line 167: `swap(lhs.adven, rhs.adven);`
Lines 175-177: 
```
for(size_t i = 0; i < lhs.adven.size(); i++) {
    swap(*lhs.adven[i], *rhs.adven[i]);
}
```

The adventurers are swapped back and forth in swap(cParty) !!

NQNStudios:
oh, whoops, I force-pushed the other PR along with this one. Redoing

CelticMinstrel:
Okay, so one is the move constructor and the other is the assignment operator…

NQNStudios:
I think I understand why the assignment operator and move constructor are both being used here.

std::move() casts univ to cUniverse&& so it qualifies for the move constructor which returns cUniverse (as a temporary value).

Then the compiler has an assignment of cUniverse& = cUniverse. It resolves that to the operator= function.

swap() is therefore called twice and I think that might actually be correct?

The real problem was double-swapping the party.adven array, I think.

CelticMinstrel:
Right, and the other thing is that the assignment operator takes its argument by value. So, in the move constructor, it should be swapping `univ` with a temporary, and then in the assignment operator it should be swapping that temporary with `real_univ`.

It is a bit silly that it swaps twice, but that doesn't appear to be the source of the issue. After all…

> Note that when I replaced `real_univ = std::move(univ)` with `swap(real_univ, univ)`, I still got the bug.

PR #333, opened 2023-01-29
vcxproj Copy scenarios and data to the right output dir

### LINKED ISSUE #332: Windows Visual Studio 2017 project: Scenarios, Base, and data copied to wrong directory
https://github.com/calref/cboe/blob/8f6450417bcce6fec6f4e09448cf0fd93385ca95/proj/vs2017/Common/Common.vcxproj#L538

This section of Common.vcxproj is meant to copy resource files into Blades of Exile Scenarios/, Blades of Exile Base/, and data/ in the output directory. But actually the build ends up with a directory structure like this:

```
cboe/
  Debug/
    Blades of Exile Scenarios/
    Blades of Exile Base/
    data/
 x64/
    Debug/
      <all the binaries>
```

Consequently the game can't load the 3 built-in scenarios.

It's because `$(SolutionDir)$(Configuration)` is used as a macro to find the output dir. When I change those to `$(SolutionDir)$(PlatformShortName)\$(Configuration)` it fixes it. PR incoming.

NQNStudios:
Fix #332 

Weirdly, things seem to run fine WITHOUT copying the data/ subfolders into the output directory. Does the resource manager embed them in the binary or something? Anyway, I made it copy them over anyway just in case.

### DIFF
--- a/proj/vs2017/Common/Common.vcxproj
+++ b/proj/vs2017/Common/Common.vcxproj
@@ -539,53 +539,53 @@
   <Target Name="CopyFiles" AfterTargets="BuildScenarios">
     <!--CURSORS-->
     <Message Text="Copying data\cursors to output directory..." />
-    <Copy SkipUnchangedFiles="true" SourceFiles="@(cursors)" DestinationFiles="@(cursors->'$(SolutionDir)$(Configuration)\data\cursors\%(RecursiveDir)%(Filename)%(Extension)')">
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
> Weirdly, things seem to run fine WITHOUT copying the data/ subfolders into the output directory.

It presumably depends on how you're launching the game. In particular, when launching from Visual Studio, it likely uses something like the repository root (or possibly the actual directory it used to copy to) as the working directory.

I hope with this it'll still launch when the files don't exist in the directory it used to copy to? If that's the case I can merge as soon as CI passes.

> Does the resource manager embed them in the binary or something?

It does not. The only things embedded in the binary are the icon and the menu definitions.

NQNStudios:
Just checked, and it does work, both launched from VSCode and by launching the generated exe directly.

PR #344, opened 2023-02-03
Change -o0 to -O0

### LINKED ISSUE #343: -o0 in CCFLAGS breaking scons debug build on linux
`scons debug=true` on Linux isn't working because it adds `-g -o0` to the compiler flags. `-o0` should be interpreted by g++ and clang as disabling optimization but it's actually being processed as setting the output file to 0. So every file is getting compiled to `cboe/0` and of course that's no good.

https://github.com/calref/cboe/blob/1bc7524ec218782ee469b2cb853f570a3f4eedcc/SConstruct#L50

Changing the flag here to `-O0` (capital O) fixes it for me.

NQNStudios:
I've only tested this change on Linux. Does CI run a debug build? If so, it'll determine whether this change works on other platforms.

Fixes #343

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -47,7 +47,7 @@ env.VariantDir('#build/obj/test', 'test')
 env.VariantDir('#build/obj/test/deps', 'deps')
 
 if env['debug']:
-   env.Append(CCFLAGS=['-g','-o0'])
+   env.Append(CCFLAGS=['-g','-O0'])
 
 # This command generates the header with git revision information
 def gen_gitrev(env, target, source):


CelticMinstrel:
> Does CI run a debug build?

That's… a good question? I think it doesn't, but presumably you can examine the CI scripts to check.

NQNStudios:
Looking at the github action logs, here's what I learned:

* None of them pass debug=true to scons. And I think that's a good thing because ultimately we want to know whether the release build works, and we don't want to wait for it to build one of each.
* Win scons is actually failing silently before it even starts a build: https://github.com/calref/cboe/actions/runs/4088304638/jobs/7051667263#step:5:11
  * It's failing to source vcvarsall.bat
  * Was win-scons even enabled in the actions until very recently?

Should I figure out what flag the Windows+Cl compiler uses for debug and add it in an if windows statement? I might not be able to test scons on Windows, I haven't gotten it working locally yet.

CelticMinstrel:
> Was win-scons even enabled in the actions until very recently?

It's been enabled for quite awhile. I'm pretty sure it was even working when the MSVC build was failing.

> Should I figure out what flag the Windows+Cl compiler uses for debug and add it in an if windows statement?

It might be nice but I think it's not crucial to have. Note that "if windows" is not enough – you need to check for the msvc toolchain as well. If it were Windows+MinGW then that uses GCC and the flags are the same as other platforms.

> I might not be able to test scons on Windows, I haven't gotten it working locally yet.

That might be because of the `vcvarsall.bat` nonsense to bring the build tools into `%PATH%`. The easiest way to make that work if building manually would probably be to use a "Visual Studio Command Prompt", which basically sets that all up for you so that all the build tools are already in `%PATH%`. It can be launched from inside Visual Studio from one of the menus. There may also be a shortcut to launch one without opening Visual Studio.

> But we should manually test debug=true for scons on every platform

scons currently doesn't work at all for me on the Mac, though I think that's only because I have SFML installed in framework format and it expects bare dynamic libraries. I suppose I could try installing those (even though it's kinda redundant) and see if it works that way. It would be kind of nice if this was a configurable option in scons but actually making that so doesn't feel worthwhile when no-one is even using scons on a Mac.

CelticMinstrel:
By the way, I believe the MSVC Cl flags equivalent to `-g -O0` would be `/Zi /Od`.

Issue #330, opened 2023-01-28
No longer builds after scons -c


x-qq:
Running `scons -c` inside the git checkout directory where build worked before somehow leads to this:
```
% scons
scons: Reading SConscript files ...                                           
Building for: posix
Using toolchain: default
C++ compiler: g++
Checking whether the C compiler works... no
There's a problem with your compiler!
```

Removing .sconsign.dblite seems to fix it.

NQNStudios:
I'm able to reproduce this.

Here's what I dug up about .sconsign.dblite:

> The .sconsign.dblite file is a temporary database used by SCons to keep file signatures to speed up future builds. If you delete it, SCons will recreate it (and your next build might take slightly longer because of that). It should not be put under version control.

(from https://stackoverflow.com/a/12867804)

So, SConstruct is marking that file for deletion, and it's getting recreated after `scons -c` is called.

Weirdly, `scons -c` still creates this bug, even after I tried changing SConstruct to not mark the dblite file for deletion.

NQNStudios:
The conversation [here](https://github.com/SCons/scons/issues/2268) indicates that you're not supposed to delete .sconsign.dblite from your SConstruct file, but `scons -c` breaks even if we don't do that, so I tried the weird `atexit` workaround shown in that issue, and then it worked.

PR #349, opened 2024-05-25
Delete .sconsign.dblite at exit of SConstruct

NQNStudios:
This fixes issue #330 for me, though I don't know if it's good scons practice or not.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1,6 +1,7 @@
 import os.path as path
 import os
 import subprocess
+import atexit
 
 # Build options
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
Looks really weird, but I don't particularly care whether it's "good scons practice" as long as it works.

I do notice that this changes it from being deleted on clean to being deleted after every build. Is that intentional?

NQNStudios:
This was working last night but now it's not working today :woman_facepalming: 

I'm going to try to find the actual root of the problem.

Issue #350, opened 2024-05-25
Dialog scale option


NQNStudios:
I can scale the UI and the map, but not the popup dialogs. On a modern screen, that text is pretty small/eye-straining.

CelticMinstrel:
Probably the main UI scale should just be applied to dialogs too.

NQNStudios:
Closing unless I get annoyed with the current solution and decide to revisit in a sophisticated way.

CelticMinstrel:
To be clear, this was implemented in #351 and is thus done now… though you're welcome to find a better way to do it at some point, I don't think it's necessary. Pixellated text is kind of expected in Blades of Exile, in my opinion. (We'd ideally disable anti-aliasing on the text, but I'm not sure how to do that.)

PR #351, opened 2024-05-27
naively scale dialogs by ui scale

NQNStudios:
This is a naive solution to #350, dialog scaling. It just scales up the dialog after drawing text (so it looks blurry), and divides mouse coordinates by scale so you can still click on everything.

It could definitely have bugs in edge cases of subclasses of Dialog, or if any dialogs don't use the Dialog class, some windows won't be fixed.

### DIFF
--- a/src/dialogxml/dialogs/dialog.cpp
+++ b/src/dialogxml/dialogs/dialog.cpp
@@ -407,7 +407,6 @@ void cDialog::loadFromFile(const DialogDefn& file){
 	dialogNotToast = true;
 	if(bg == BG_DARK) defTextClr = sf::Color::White;
 	// now calculate window rect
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
Looks fine to me, though you may want to consider putting `get_float_pref("UIScale", 1.0)` into a private function in cDialog, since it's used at least 3 times.

Doesn't the main UI scaling also scale up after text? It should end up being pixellated rather than blurry, though.

CelticMinstrel:
> It could definitely have bugs in edge cases of subclasses of Dialog, or if any dialogs don't use the Dialog class, some windows won't be fixed.

I don't think there are any that don't use the Dialog class, and as for subclasses, I'm pretty sure there aren't any.

PR #356, opened 2024-05-30
Local build improvement

NQNStudios:
I went through all of @clort81's challenges when building the latest source code, and added automation/documentation for everything with a known solution.

* Catch2 is now a *subtree*, not a *submodule*, because [subtrees](https://www.atlassian.com/git/tutorials/git-subtree) automatically clone without `git clone --recursive` or `git submodule update --init`
* Added TGUI 0.9 as a subtree, and updated the Scons build to build it from source using C++14 if the library isn't detected installed on the system.
* Updated README.md to explain the annoying particularities of installing TGUI if folks choose to do it themselves


NQNStudios:
I tested it locally on a fresh clone and actually it needs a couple more things.

NQNStudios:
Ok, just tested it and `scons` works on the first try on a fresh clone on linux now, which was the goal.

NQNStudios:
Ok, I just tested it locally and a fresh clone on linux can just run `scons` as long as they have all the dependencies, plus `cmake` to build TGUI

CelticMinstrel:
> * Catch2 is now a _subtree_, not a _submodule_, because [subtrees](https://www.atlassian.com/git/tutorials/git-subtree) automatically clone without `git clone --recursive` or `git submodule update --init`

I'm a bit unsure about this part…

NQNStudios:
It's basically the same as a submodule, but we'll never have to remind anyone to call `submodule update --init`. I pulled the same version of Catch2 as we were using before.

If Catch2 were an optional dependency I could see an advantage to not always cloning it. But currently we have a state where everyone who tries to build from a normal git clone will get an easily avoidable compiler error--I don't see the point.

CelticMinstrel:
After reading up on subtree and checking why I made Catch a submodule in the first place, I'm even less sold on the idea of the subtree. If the only issue is that submodules aren't automatically cloned, I think I'd rather just add a step in scons that makes sure the submodules are present before building things that depend on them. In other words, have scons call `git submodule update --init` for you.

And I also notice that you made TGUI a subtree here. I think linking it in as a submodule would actually be a good idea.

NQNStudios:
People building with Xcode or Visual Studio will still fall into the trap
of Catch2 being missing, though

CelticMinstrel:
That's true, but people building with Xcode or Visual Studio would still be able to build the game even if Catch is missing. They'd only get a build failure if they try to build the tests. (This _should_ be true of scons too, I just haven't bothered to work out how to make scons have multiple targets.)

PR #357, opened 2024-05-31
Local build improvement redo

NQNStudios:
This is the last PR, with submodules instead of subtrees, and a few more things:

* add Catch2 dependency to the README
* partial build options
* correct debug flags on all platforms
* when scons -c is called, print a note about deleting .sconsign.dblite (this addresses #330, which I still can't reliably fix, but deleting the file manually works every time)

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -70,7 +70,10 @@ env.VariantDir('#build/obj/test', 'test')
 env.VariantDir('#build/obj/test/deps', 'deps')
 
 if env['debug']:
-   env.Append(CCFLAGS=['-g','-o0'])
+	if platform in ['posix', 'darwin']:
+		env.Append(CCFLAGS=['-g','-O0'])
+	elif platform == 'win32':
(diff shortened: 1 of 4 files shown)


SConstruct
```diff
@@ -221,12 +248,15 @@ if platform == 'darwin':
...
 	if platform == 'darwin':
-		env.Append(FRAMEWORKPATH=['deps/lib'])
+		env.Append(FRAMEWORKPATH=[os.getcwd() + '/deps/lib'])
+
+if path.exists('deps/lib64'):
+	env.Append(LIBPATH=[os.getcwd() + '/deps/lib64'])
 
 if path.exists('deps/include'):
-	env.Append(CPPPATH=['deps/include'])
+	env.Append(CPPPATH=[os.getcwd() + '/deps/include'])
```
CelticMinstrel, line 259: I'm not really a fan of concatenating `os.getcwd()` like this. I'd prefer to use a dedicated path join function (either `os.path` or `pathlib`).

PR #359, opened 2024-06-11
Fix win scons

NQNStudios:
There's some messiness in this PR with trying things, reverting them, etc. It should probably all be squashed, but I can now make a build with scons on Windows that actually launches.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -321,6 +321,7 @@ if not env.GetOption('clean'):
 
 	boost_versions = ['-1_84'] # This is a bit of a hack. :(
 	msvc_versions = ['140', '141', '142', '143'] # This is a new bit of a hack. :(
+	suffixes = ['-mt', '-mt-x64', '-mt-x32']
 	bundled_libs = []
 
 
@@ -330,8 +331,8 @@ if not env.GetOption('clean'):
 	check_header('boost/any.hpp', 'Boost.Any')
 	check_header('boost/math_fwd.hpp', 'Boost.Math')
 	check_header('boost/spirit/include/classic.hpp', 'Boost.Spirit.Classic')
-	check_lib('boost_system', 'Boost.System', ['-mt'], boost_versions, msvc_versions)
-	check_lib('boost_filesystem', 'Boost.Filesystem', ['-mt'], boost_versions, msvc_versions)
+	check_lib('boost_system', 'Boost.System', suffixes, boost_versions, msvc_versions)
+	check_lib('boost_filesystem', 'Boost.Filesystem', suffixes, boost_versions, msvc_versions)
 	check_lib('sfml-system', 'SFML-system')
 	check_lib('sfml-window', 'SFML-window')
 	check_lib('sfml-audio', 'SFML-audio')
--- a/SConstruct
+++ b/SConstruct
@@ -372,6 +372,8 @@ env.Append(CPPDEFINES=["TIXML_USE_TICPP"])
 if platform == "win32":
 	# For the *resource.h headers
 	env.Append(CPPPATH=["#rsrc/menus"])
+	# Icons
+	env.Append(CPPPATH=["#rsrc/icons/win"])
 
 if platform == "darwin":
 	env.Append(LIBS=Split("""
--- a/SConstruct
+++ b/SConstruct
@@ -190,13 +190,16 @@ elif platform == "win32":
 		vcpkg_other_packages = Glob(path.join(vcpkg_prefix, 'packages', f'**x{env["bits"]}-windows'))
 		vcpkg_other_includes = [path.join(d.get_abspath(), 'include') for d in vcpkg_other_packages]
 		vcpkg_other_libs = [path.join(d.get_abspath(), 'lib') for d in vcpkg_other_packages]
-		project_includes = [path.join('src', dir) for dir in filter(lambda dir: os.path.isdir(path.join('src', dir)), os.listdir('src'))]
+		project_includes = []
+		for (root, dirs, files) in os.walk('src'):
+			project_includes.append(path.join(os.getcwd(), root))
+
+		# project_includes = [path.join('src', dir) for dir in filter(lambda dir: os.path.isdir(path.join('src', dir)), os.listdir('src'))]
 		include_paths=[path.join(vcpkg_installed, 'include')] + vcpkg_other_includes + project_includes
 		env.Append(
 			LINKFLAGS=['/SUBSYSTEM:WINDOWS','/ENTRY:mainCRTStartup',f'/MACHINE:X{env["bits"]}'],
 			CXXFLAGS=['/EHsc','/MD','/FIglobal.hpp'],
-			CPPPATH=[path.join(vcpkg_installed, 'include')],
-			INCLUDEPATH=include_paths,
+			CPPPATH=include_paths,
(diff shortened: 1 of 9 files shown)


SConstruct
```diff
@@ -185,12 +185,25 @@ if platform == "darwin":
...
+		vcpkg_other_bins = [path.join(d.get_abspath(), 'bin') for d in vcpkg_other_packages]
+		vcpkg_other_bins = list(filter(path.exists, vcpkg_other_bins))
+		project_includes = []
+		for (root, dirs, files) in os.walk('src'):
+			project_includes.append(path.join(os.getcwd(), root))
+
+		include_paths=[path.join(vcpkg_installed, 'include')] + vcpkg_other_includes + project_includes
 		env.Append(
-			LINKFLAGS=['/SUBSYSTEM:WINDOWS','/ENTRY:mainCRTStartup','/MACHINE:X86'],
+			LINKFLAGS=['/SUBSYSTEM:WINDOWS','/ENTRY:mainCRTStartup',f'/MACHINE:X{env["bits"]}'],
```
CelticMinstrel, line 203: Does `/MACHINE:X32` actually work though? Because that's what this is going to pass in if you try to do a win32 build.

SConstruct
```diff
@@ -312,8 +326,8 @@ if not env.GetOption('clean'):
 			print("  If you're sure it's installed, try passing INCLUDEPATH=...")
 			Exit(1)
 
-	boost_versions = ['-1_54', '-1_55', '-1_56', '-1_57', '-1_58'] # This is a bit of a hack. :(
-	bundled_libs = []
+	boost_versions = ['-1_84'] # This is a bit of a hack. :(
+	suffixes = ['-mt', f'-mt-x{env["bits"]}']
```
CelticMinstrel, line 330: Again, is `-mt-x32` a valid suffix?

SConstruct
```diff
@@ -447,22 +472,23 @@ elif platform == "win32":
...
+			os.makedirs("build/Blades of Exile", exist_ok=True)
 			open("build/Blades of Exile/VCRedistInstall.exe", 'w').close()
 
 if platform == "darwin":
 	env.VariantDir("#build/pkg", "pkg/mac")
 	SConscript("build/pkg/SConscript")
-elif platform == "win32" and subprocess.call(['where', '/Q', 'makensis']) == 0:
-	env.VariantDir("#build/pkg", "pkg/win")
-	SConscript("build/pkg/SConscript")
+#elif platform == "win32" and subprocess.call(['where', '/Q', 'makensis']) == 0:
```
CelticMinstrel, line 489: This needs to be uncommented or otherwise fixed before I can merge it.

src/game/boe.menus.win.cpp
```diff
@@ -4,13 +4,13 @@
 #include <sstream>
 #include <SFML/Graphics/RenderWindow.hpp>
 #include "boeresource.h"
-#include "universe.hpp"
+#include "../universe/universe.hpp"
```
CelticMinstrel, line 7: Don't use include paths relative to the current directory, especially not paths containing `..`.

SConstruct
NQNStudios: The installer part is having weird problems. I didn't want to deal with it because I think an auto-updating deployment target like Itch or Steam is preferable to having an installer at all.

SConstruct
```diff
@@ -422,14 +438,23 @@ if platform == "darwin":
...
 		zlib1
+		freetype
+		vorbis
+		vorbisfile
+		vorbisenc
+		ogg
+		FLAC
+		bz2
+		brotlidec
+		brotlicommon
```
CelticMinstrel, line 457: Something seems to be wrong here. What on earth is "brotli", and why is bz2 being bundled when we don't use it?

I don't think we use vorbis/ogg/FLAC, either, though it wouldn't hurt to include it for use by mods someday. Including freetype is fine.

NQNStudios: I agree it seems wrong, but the Visual Studio build outputs all of those DLLs, and the scons build won't launch without them bundled.

SConstruct
CelticMinstrel: Putting it on Itch or Steam is fine, but we should support a standalone installer build as well to support people who don't like using services like Itch or Steam. What kind of weird problems was it having? Maybe there's an issue of using deprecated NSIS stuff or something…

SConstruct
CelticMinstrel: What do you mean it "outputs" all of those DLLs?

src/game/boe.menus.win.cpp
NQNStudios: I left these in by accident. I put them back to normal and my local build still worked so hopefully the CI re-run does too.

SConstruct
NQNStudios: They're all in the build directory Visual Studio outputs, and if you rename one of them to something else, the binary won't launch.

SConstruct
NQNStudios: As it turns out, yes. One place where it's 32 and not 86.

SConstruct
CelticMinstrel: So then the question is, what exactly is causing them to be required? They shouldn't be required, since we don't use them anywhere. Are they transitive dependencies of something else? If so, the something else should be built in a way that doesn't require them.

NQNStudios:
I can't get the 32-bit build to work locally despite quadruple-checking all the lib paths. The compiler isn't finding Zlib, but I'm 100% sure I have an x86 build of zlib installed where scons is checking for it. This is frustrating.

I want to suggest separating out some concerns.

Without the PR, we have these problems:
- [ ] impossible to make a 32-bit build with scons on windows
- [ ] impossible to make a 64-bit build with scons on windows
- [ ] no installer generated by any Windows build
- [ ] windows scons logic missing so many required steps, I struggle to imagine how it ever worked, and/or what manual steps were required locally to make it work, but aren't documented anywhere
- [ ] unwanted DLLs required by the windows builds

With the PR, the situation becomes:

- [x] impossible to make a 64-bit build with scons on windows
- [x] windows scons logic missing so many required steps, I struggle to imagine how it ever worked, and/or what manual steps were required locally to make it work, but aren't documented anywhere
- [ ] impossible to make a 32-bit build with scons on windows
- [ ] no installer generated by any Windows build
- [ ] unwanted DLLs required by the windows builds

I would suggest that we do this:

1. Open an issue for the win-scons 32-bit build
2. Open an issue for the win-scons installer generation
3. Open an issue for the unwanted DLLs
4. Set the default win-scons build to 64-bit, so contributors following the README should get a successful result by default
5. Merge the PR

CelticMinstrel:
> I struggle to imagine how it ever worked

I'm not certain that it ever worked. When I built on Windows, I (almost?) always used MSVC 2013, not scons.

I agree on points 1, 3, and 4. For point 2, I would like the installer generation uncommented, but it can be set to run only if explicitly requested (so, running `scons` won't try to generate an installer, but running something like `scons package` will).

NQNStudios:
I did all the things.

CelticMinstrel:
Is it okay to squash all of this into a single commit? And if so, do you want to write up a summarized commit message?

NQNStudios:
I'm pretty comfortable squashing all of this one without doing a summary. 🤷‍♀️

PR #365, opened 2024-06-13
[WIP] Replay system

NQNStudios:
This is the big one.


src/tools/prefs.win.cpp
NQNStudios, line 1: Preferences need to be recorded and retrieved by the replay system, or else the replays won't be deterministic in a lot of very unpredictable ways.

On Mac, a snazzy system library is used to manage preferences. But I think we should change things so that Mac handles preferences the same as Windows and Linux, because that way recording and replay can be deterministic on Mac too.

CelticMinstrel: I'd prefer not to change how the Mac stores preferences, and I don't understand how the Mac's preferences system gets in the way of what you want to do.

src/tools/prefs.win.cpp
```diff
@@ -99,10 +100,18 @@ static bool save_prefs(fs::path fpath) {
 static bool load_prefs(fs::path fpath) {
 	prefsDirty = false;
 	std::map<std::string,boost::any> temp_prefs;
-	std::ifstream fin(fpath.string().c_str());
+	std::istream* in;
```
CelticMinstrel, line 108: Please don't use raw pointers. This should probably be an `std::unique_ptr`.

proj/xc12/BoE.xcodeproj/project.pbxproj
```diff
@@ -325,6 +325,7 @@
 		91FCC8F418FF0866007026CE /* pc.appleevents.mm in Sources */ = {isa = PBXBuildFile; fileRef = 91FCC8F318FF069A007026CE /* pc.appleevents.mm */; };
 		91FD417423DFC9D4003D5B8A /* boe.ui.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 91FD417223DFC5C1003D5B8A /* boe.ui.cpp */; };
 		91FE0E3823F084B70084CA6B /* drawable_manager.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 91FE0E3523F084B70084CA6B /* drawable_manager.cpp */; };
+		D384F6A32C1B9D4000A806C3 /* replay.cpp in Sources */ = {isa = PBXBuildFile; fileRef = D384F6A22C1B9D4000A806C3 /* replay.cpp */; };
```
CelticMinstrel, line 328: Did you do this in Xcode, with a tool, or manually?

src/tools/prefs.win.cpp
NQNStudios: So, looking through the code it seems like a Mac library handles all of the preference syncing behind the scenes, and I couldn't find an API function that would let me iterate through and serialize all of the preferences. It looked like the Mac API provides a dictionary object where you have to know a key to get it's value.

Which is really, really, weird. So maybe I just missed something.

src/tools/replay.cpp
```diff
@@ -0,0 +1,78 @@
...
+		} catch(...) {
+			std::cout << "Failed to load file " << file << std::endl;
+		}
+		return true;
+	}
+	return false;
+}
+
+void record_action(std::string action_type, std::string inner_text) {
+	Element* root = log_document.FirstChildElement();
```
CelticMinstrel, line 57: This isn't really an issue, but I'm noticing you didn't use the XML printer class that the scenario editor uses to write scenario files. It might not actually be appropriate here, I'm not sure; but in case you simply didn't know about it, I thought I'd mention it.

CelticMinstrel (commented):
Looks decent enough as a starting point. The goal of recording the _entire application_ (as opposed to _just_ a game session) seems slightly absurd to me, but I guess we'll see.

src/tools/prefs.win.cpp
NQNStudios: https://github.com/calref/cboe/blob/master/src/tools/prefs.mac.mm#L73

Under the hood, MacOS is using this NSUserDefaults object to store preferences: https://developer.apple.com/documentation/foundation/nsuserdefaults

NQNStudios: Oh. This might be what I was looking for and couldn't find: https://developer.apple.com/documentation/foundation/nsuserdefaults/1415919-dictionaryrepresentation

NQNStudios: So maybe I can use that function to write Mac-specific logic that serializes and deserializes the NSUserDefaults, but that seems weird to me, when we could run the same code on all 3 platforms instead.

proj/xc12/BoE.xcodeproj/project.pbxproj
NQNStudios: In Xcode.

src/tools/prefs.win.cpp
NQNStudios: I also don't know the programming language that `prefs.mac.mm` is written in, which is awkward. Is it Objective-C?

src/tools/prefs.win.cpp
NQNStudios: Should I also change the method signature of pop_next_action() to return `unique_ptr<Element>`  instead of `Element*`? Seems like that would be good, because I realized I'm cloning Elements and passing the clones out of pop_next_action() but not deleting them ever.

src/tools/prefs.win.cpp
CelticMinstrel: Yes and no. It's Objective-C++ – an unholy fusion of C++ and Objective-C.

src/tools/replay.cpp
```diff
@@ -62,7 +62,7 @@ void record_action(std::string action_type, std::string inner_text) {
 	log_document.SaveFile(log_file);
 }
 
-Element* pop_next_action(std::string expected_action_type) {
+std::unique_ptr<Element> pop_next_action(std::string expected_action_type) {
```
CelticMinstrel, line 73: I don't really understand why this became a `unique_ptr`. I know I said not to use raw pointers, but the pointers to XML objects are pointing to persistent memory inside the XML document, right? So I think they could just be passed around as references.

…though I guess that would assume you're not modifying the document as you go.

But is making it a `unique_ptr` like this actually safe? I don't know what `Clone()` did to allocate this element.

src/tools/prefs.win.cpp
CelticMinstrel: Something like that is definitely needed. I'm not sure if that exact thing is it. It all depends on how `Clone()` works.

src/game/boe.main.cpp
```diff
@@ -250,6 +250,15 @@ void process_args(int argc, char* argv[]) {
 	}
 }
 
+void replay_next_action() {
+	auto next_action = pop_next_action();
+	std::string t = next_action->Value();
+	if (t == "load_party") {
+		decode_file(next_action->GetText(), "temp.exg");
+		load_party("temp.exg", univ);
```
CelticMinstrel, line 258: This seems a bit sloppy. Where is that file placed on disk? I would guess it ends up in the current working directory, which is probably right beside the executable, but that's a problem, because it's usually write-protected.

I'd say it should be put in the same place that temporary scenario files are written to when editing a packed scenario. I think that was something like a "temp" folder within the settings folder.

src/tools/replay.cpp
NQNStudios: It was, indeed, wrong to use `unique_ptr` and causing a crash that I couldn't figure out last night. I'm not very well versed in the standard library pointer types.

The behavior I wanted was this:

* there is one global variable, log_document, which loads all the xml at launch and unloads it at program exit
* pop_next_action() removes the next element from log_document, passing ownership of the element's data to the calling context
* the element's data is released when the calling context drops it.

I think the actual way to achieve this is to return the `Element` plainly. The calling context allocates it on the stack, then cleans it up.

CelticMinstrel: Could be, yes.

src/tools/replay.cpp
```diff
@@ -83,8 +83,7 @@ std::unique_ptr<Element> pop_next_action(std::string expected_action_type) {
 	Element* clone = next_action->Clone()->ToElement();
 	root->RemoveChild(next_action);
 	
-	std::unique_ptr<Element> ptr { clone };
-	return ptr;
+	return *clone;
```
CelticMinstrel, line 86: If this is what you're doing, do you really need to clone? Couldn't you just replace the clone line by `Element clone = *next_action;`?

NQNStudios: Calling RemoveChild() on next_action was freeing the data. Cloning is necessary to keep it around without leaving it in the document object

CelticMinstrel: Are you saying that `Element clone = *next_action` does not copy it?

But as it is now, isn't the clone still being leaked?

NQNStudios: when a block does

```
{
   Element next = pop_next_action()
}
```

I thought the block is taking ownership of the data, then when the scope ends, it gets deleted.

NQNStudios: > Are you saying that Element clone = *next_action does not copy it?

Maybe it would.

CelticMinstrel: > I thought the block is taking ownership of the data, then when the scope ends, it gets deleted.

That's correct, but I'm talking about the clone _inside_ the function. That's a pointer, so it's not going to be automatically deleted.

NQNStudios: I think I get what you're saying. Let me fix that, then

NQNStudios: For it to not crash, Clone() is required.

`Element clone = *(next_action->Clone()->ToElement());`

This works. Does that avoid a memory leak?

CelticMinstrel: I think calling `Clone()` in and of itself _might_ be a memory leak, but I'd have to see the internals of `Clone()` be sure. The _typical_ internals of a function called clone is essentially `return new ClassName(*this)`, and if that's what it's doing, then this line is certainly a memory leak. But maybe `Clone()` is actually doing something else. I'm not sure.

NQNStudios: I looked into Clone() and it's definitely a memory leak. Which brings me full-circle to what I was trying to accomplish by returning a `unique_ptr`: I thought wrapping the pointer returned by Clone() in unique_ptr would make the `unique_ptr` manage that memory, so when the pointer went out of scope in the calling context of `pop_next_action()`, it would delete the clone.

NQNStudios: I think the document object might still manage the memory of the pointer returned by `Clone()` though. Because I get a segfault when `unique_ptr` manages the pointer and frees it.

CelticMinstrel: > I looked into Clone() and it's definitely calling `new`. Which brings me full-circle to what I was trying to accomplish by returning a `unique_ptr`: I thought wrapping the pointer returned by Clone() in unique_ptr would make the `unique_ptr` manage that memory, so when the pointer went out of scope in the calling context of `pop_next_action()`, it would delete the clone.

This is a more or less correct line of reasoning. The only problem is, it's not clear whether deleting the clone is the correct way to dispose of it.

NQNStudios:
That's annoying -- I wanted to rebase this onto master and it ended up adding commits from master into the PR

CelticMinstrel:
Kinda looks like you rebased it backwards? If you're on the replay branch, you want to rebase it onto master with `git rebase master`. You definitely _don't_ want to be on master and rebase onto this branch.

src/tools/replay.cpp
CelticMinstrel: What's the situation with this? Is it appropriate to use my XML printer class here or not?

Issue #59, opened 2017-09-05
A strip of pattern along the bottom of the window doesn't update


CelticMinstrel:
This occurs on Windows in any fullscreen mode.

NQNStudios:
@CelticMinstrel I fixed this in the most recently merged PR but it didn't get automatically closed, could you close it please? :+1:

CelticMinstrel:
Right, this was fixed in #367

PR #367, opened 2024-06-19
Fix graphics flickering/stretching on Windows

### LINKED ISSUE #366: [Windows] Bad graphical flickering in Windowed mode
cboeGif

Windowed mode is screwy on Windows. I tried it with and without UI scaling. This also affects the scenario and character editors.

The gif I recorded doesn't capture the speed at which the flickering really happens, which is actually much faster.

NQNStudios:
Fixes #366 

This bug was unbelievably hard to track down and fix.

I think it was probably caused by a change in SFML's code for handling window resize events on Windows. When we add the menubar to the window by bypassing SFML and using the Windows APIs, SFML now detects that change and shrinks the logical viewport of your window without expanding the window itself to fit the menubar in it.

I updated `menuProc()` to catch that first unwanted resize event and discard it. I also made the window expand to fit the menubar's additional height, handled for a rare edge case where the menubar needs more than 1 row for its buttons, and made sure the splash screens still render in the right places.

I need to test this on Linux and Mac to make sure it didn't break those.

### DIFF
--- a/src/game/boe.startup.cpp
+++ b/src/game/boe.startup.cpp
@@ -112,16 +112,24 @@ void handle_splash_events() {
 	while (mainPtr.pollEvent(event)) {
 		if (event.type == sf::Event::GainedFocus || event.type == sf::Event::MouseMoved)
 			set_cursor(sword_curs);
+		if (event.type == sf::Event::Resized) {
+			std::cout << "new width: " << event.size.width << std::endl;
+			std::cout << "new height: " << event.size.height << std::endl;
+		}
 	}
 }
 
+rectangle view_rect() {
+	sf::Vector2f size = mainPtr.getView().getSize();
+	return rectangle(0, 0, size.y, size.x);
+}
+
 void show_logo() {
-	rectangle whole_window;
+	rectangle whole_window = view_rect();
 	
 	if(get_int_pref("DisplayMode") != 5)
 		hideMenuBar();
 	
-	whole_window = rectangle(mainPtr);
 	double ui_scale = get_float_pref("UIScale", 1.0);
 	if(ui_scale < 1) ui_scale = 1;
 	rectangle logo_from = {0, 0, int(ui_scale *350), int(ui_scale * 350)};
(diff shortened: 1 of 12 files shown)


NQNStudios:
This doesn't break the Linux behavior.

src/game/boe.startup.cpp
```diff
@@ -109,9 +109,10 @@ bool handle_startup_press(location the_point) {
 
 void handle_splash_events() {
 	sf::Event event;
-	if(!mainPtr.pollEvent(event)) return;
-	if(event.type == sf::Event::GainedFocus || event.type == sf::Event::MouseMoved)
-		set_cursor(sword_curs);
+	while (mainPtr.pollEvent(event)) {
```
CelticMinstrel, line 112: No space after while please.

src/tools/winutil.win.cpp
```diff
@@ -444,5 +444,11 @@ ModalSession::~ModalSession() {
 }
 
 int getMenubarHeight() {
-	return GetSystemMetrics(SM_CYMENU);
+	MENUBARINFO info;
+	info.cbSize = sizeof(MENUBARINFO);
+	if (GetMenuBarInfo(mainPtr.getSystemHandle(), OBJID_MENU, 0, &info))
```
CelticMinstrel, line 449: Ditto.

NQNStudios:
I checked and this also doesn't break anything on Mac OS.

NQNStudios:
These changes didn't fix the scenario and character editors, though...

NQNStudios:
Would be bad practice to leave something broken in the editors when there's a known solution, so I applied the same fix to them and it worked.

NQNStudios:
I think issue #59 must be related to the same underlying cause (stuffing a menu bar into an SFML window making it squish its effective viewport weirdly). So I might as well see if that's still a problem and  try to fix that too.

NQNStudios:
@CelticMinstrel I'm done fiddling with this one if you don't have any other changes in mind.

CelticMinstrel:
I assume you checked that this doesn't break anything on Mac or Linux, right?

NQNStudios:
Yeah I did

PR #346, opened 2023-02-04
Redo2-II

fosnola:
I just rebased redo2, so a new pull request.

**Note**: there were many conflicts to solve, I tried to solve them, but I don't guarantee anything...

PR #370, opened 2024-06-20
Fix fullscreen window sizing on MacOS HighDPI display

### LINKED ISSUE #369: [Mac] fullscreen window is dramatically larger than my screen
Something with the fullscreen handling on MacOS is weird, possibly having to do with retina displays and/or MacOS window scaling.

The splash screens render clipped off in the bottom right corner of the window. Zoom out and you can see this is because the fullscreen window is more than twice as large as it should be, not fitting onto the screen. So the splash screens are rendering correctly centered, but on a window that is way too big.

If your fullscreen preference is anchored anywhere other than the top-left, you won't be able to see or click on most of the UI, which *does* draw at the correct size for my scale setting.

Another bug which I think must be related: dialog windows open in the bottom right corner of my screen. They must be centering themselves within the much much larger rectangle of the main window.

NQNStudios:
Fixes #369 in a counterintuitive way.

I would have thought that we would want the flag that says High DPI displays *are* supported (which was even something [Fosnola](https://github.com/calref/cboe/pull/346/commits/ff05298ebec161cded8b8aa0821b4879bd0b9954) had done and I tried cherry-picking to fix this.)

But this [very old SFML forum thread](https://en.sfml-dev.org/forums/index.php?topic=17241.msg124542#msg124542) suggested setting the very same flag to false. Which I did, and it fixed both parts of the issue--the fullscreen window size, and the positioning of dialog windows.

### DIFF
--- a/pkg/mac/BoE-Info.plist
+++ b/pkg/mac/BoE-Info.plist
@@ -239,5 +239,7 @@
 			</dict>
 		</dict>
 	</array>
+	<key>NSHighResolutionCapable</key>
+	<false/>
 </dict>
 </plist>
--- a/pkg/mac/BoECharEd-Info.plist
+++ b/pkg/mac/BoECharEd-Info.plist
@@ -69,5 +69,7 @@
 			</dict>
 		</dict>
 	</array>
+	<key>NSHighResolutionCapable</key>
+	<false/>
 </dict>
 </plist>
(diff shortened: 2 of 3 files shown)


CelticMinstrel:
I think this probably _is_ correct. After all, OBoE is indeed not conscious of hi-DPI displays, requiring the user to intervene manually to apply a scaling factor.

PR #371, opened 2024-06-20
More homebrew search paths in Mac Scons

NQNStudios:
My Mac installed homebrew libraries into /opt/homebrew, not /usr/local, so the scons build didn't work for me locally without this change. I also had to point it to sfml.


CelticMinstrel:
This is kinda annoying how Homebrew is so confusing about where it actually installs the files… that might be _part_ of why I'm still insisting on using MacPorts instead…

PR #375, opened 2024-07-03
[WIP] Replay system rebased

NQNStudios:
Rebased continuation of #365


CelticMinstrel:
It doesn't look like the cloning issue in `pop_next_action` was resolved.

NQNStudios:
I couldn't find any way to make the code not crash other than taking the raw `Element*` pointer and trusting it gets deleted somehow. So I went back to that.

I think the best solution is to actually not remove the elements, just keep a pointer to the next one, and iterate through them leaving the data structure intact. Won't need to Clone() that way, and will only work with references to the data.

Also, I finally looked at the XML printer class and I think I can use it for this.

NQNStudios:
Well, there's one hitch with the printer class. It only saves when it's deconstructed. I think saving the replay log is desirable every time an action is added, in case a segfault or exception crashes the program. We want to record as much of a playthrough as we can up until the moment of failure, not rely on the program to run and finish normally to keep the log alive.

src/tools/replay.cpp
```diff
@@ -79,32 +82,25 @@ void record_action(std::string action_type, std::map<std::string,std::string> in
 }
 
 bool has_next_action() {
-	Element* root = log_document.FirstChildElement();
-	return root->FirstChildElement(false) != NULL;
+	return next_action != NULL;
```
CelticMinstrel, line 85: Use nullptr instead of NULL, please.

CelticMinstrel:
> I couldn't find any way to make the code not crash other than taking the raw `Element*` pointer and trusting it gets deleted somehow. So I went back to that.
> 
> I think the best solution is to actually not remove the elements, just keep a pointer to the next one, and iterate through them leaving the data structure intact. Won't need to Clone() that way, and will only be returning references to the data.

Just double-checking, because I can't really tell very well from the code, but… are you using the ticpp API here rather than the TinyXML API that it wraps?

> Well, there's one hitch with the printer class. It only saves when it's deconstructed. I think saving the replay log is desirable every time an action is added, in case a segfault or exception crashes the program. We want to record as much of a playthrough as we can up until the moment of failure, not rely on the program to run and finish normally to keep the log alive.

If the printer class doesn't fit the requirements for what you're doing, then don't worry about it. I just wanted to make sure you hadn't overlooked that it exists, in case it made things easier for you. If it makes things more difficult instead, feel free to forget about it.

src/game/boe.startup.cpp
```diff
@@ -37,71 +41,81 @@ extern sf::View mainView;
 
 enum_map(eStartButton, rectangle) startup_button;
 
-// TODO: Always returns false, so make it void
-bool handle_startup_press(location the_point) {
+void handle_startup_button_click(eStartButton btn) {
+	if(recording){
+		stringstream sstr;
```
CelticMinstrel, line 46: This should be ostringstream, not stringstream.

src/fileio/fileio_party.cpp
```diff
@@ -30,6 +30,28 @@ extern cCustomGraphics spec_scen_g;
 static bool load_party_v1(fs::path file_to_load, cUniverse& univ, bool town_restore, bool in_scen, bool maps_there, bool must_port);
 static bool load_party_v2(fs::path file_to_load, cUniverse& univ);
 
+fs::path nav_get_party();
```
CelticMinstrel, line 33: If you must declare these two functions here in this file instead of including a header, please make the following two changes:

1. Move them both to the top of the file (that is, the second one should be immediately below this one instead of further down).
2. Declare them extern.

CelticMinstrel:
What's the status of this? It's still marked as a draft, but it looks almost mergeable (as long as we can prove the Element pointers aren't being leaked).

NQNStudios:
Thanks for checking in. I’ve been banging my head against the wall trying
to do automated release builds in a branch of my fork. It might be good to
take a break from that and come back to this branch.

Are you willing to merge this PR with only the current subset of input
actions supported in the record/replay system? I was thinking I’d keep it a
draft until most everything is implemented. But merging what’s there now
could be good too so that the bug fix PRs I keep making along the way don’t
continue to make rebasing the replay PR harder.

I fixed the Element* memory problem by no longer removing elements from the
document object. Instead of actually “popping”, it’s more like iterating
now. So cloning isn’t required. I left the function name as
pop_next_action() because “popping” the event log from head to tail is
still the spirit of the abstraction although it is not literally the
underlying implementation.

CelticMinstrel:
This more or less sets up the basic system, so it's in a state where I'd be fine merging it, as long as you've also addressed those two comments above – the function prototypes and the stringstream.

NQNStudios:
#379 is this rebased

PR #376, opened 2024-07-04
Fix some repetitive code in dialogs

NQNStudios:
I saw a couple of repetitive things in cControl and cDialog, one of which was marked with a TODO to deduplicate it.

In the first case, I simply made a function and changed the 2 repetitive places to call the function.

The second case was more complicated, and my change actually does change the behavior, but I don't think it should matter unless any widgets have containers in containers and we want those *not* to be processed recursively (how it was before).


CelticMinstrel:
Took me a while to work out what the change was that you made in the second commit, but now that I've worked it out, it seems fine to me.

This is out of scope for this PR, probably, but the first commit makes me realize that it might make sense for cControl to have a member variable that specifies the click sound. Dynamic casting to `cLed` as a special case is such a hack.

NQNStudios:
The CI failure for win-vs64 looks like an intermittent unit test failure that happens sometimes.

PR #379, opened 2024-07-27
Replay system part 1

NQNStudios:
This is rebased with all the little things fixed.


src/game/boe.main.cpp
```diff
@@ -245,21 +319,38 @@ void init_boe(int argc, char* argv[]) {
...
 	set_cursor(watch_curs);
 	init_buf();
-	srand(time(nullptr));
+
+	// Seed the RNG
+	if (replaying) {
+		auto srand_element = pop_next_action("srand");
+		
+		std::string ts(srand_element->GetText());
+		srand(atoi(ts.c_str()));
```
CelticMinstrel, line 310: Seeing this made me realize that we may want to consider moving away from `rand()` in favour of a Mersenne twister… but that would be something for a separate PR.

src/tools/prefs.win.cpp
```diff
@@ -99,10 +101,18 @@ static bool save_prefs(fs::path fpath) {
 static bool load_prefs(fs::path fpath) {
 	prefsDirty = false;
 	std::map<std::string,boost::any> temp_prefs;
-	std::ifstream fin(fpath.string().c_str());
+	std::istream* in;
```
CelticMinstrel: This should be an `std::unique_ptr<std::istream>`. (If that causes problems for some reason, `shared_ptr` is also acceptable.)

src/game/boe.main.cpp
```diff
@@ -233,11 +234,18 @@ std::string log_file;
 
 bool init_action_log(std::string command, std::string file) {
 	if (command == "record") {
+		// If a filename is given, use it as a base, but insert a timestamp for uniqueness.
+		if (file.empty())
+			file = "BoE";
+
+		if (boost::ends_with(file, ".xml"))
```
CelticMinstrel: I think Boost.Algorithm is a new dependency, right? I don't mind adding it, but it needs to be mentioned in the README.

NQNStudios:
I must have missed something in the rebase, I think I had that fixed

src/tools/prefs.win.cpp
CelticMinstrel: Ah, looks like you already made that change in a later commit, so you can disregard this.

src/tools/replay.cpp
```diff
@@ -79,30 +82,25 @@ void record_action(std::string action_type, std::map<std::string,std::string> in
...
 	}
 
-	Element* clone = next_action->Clone()->ToElement();
-	root->RemoveChild(next_action);
+	Element* to_return = next_action;
+	
+	next_action = next_action->NextSiblingElement(false);
 	
-	return clone;
+	return to_return;
```
CelticMinstrel, line 103: Is it intended that pop_next_action can return null? If it's never null, it should be returned as a reference, but if the expectation is that it could return null, then this is fine as is.

src/game/boe.main.cpp
```diff
@@ -257,11 +259,15 @@ void replay_next_action() {
 
 	auto next_action = pop_next_action();
 	std::string t = next_action->Value();
-	if(t == "load_party") {
+	
+	if(overall_mode == MODE_STARTUP && t == "startup_button_click"){
+		eStartButton btn = static_cast<eStartButton>(atoi(next_action->GetText().c_str()));
```
CelticMinstrel, line 264: I'm not going to consider this a blocker, but I think it would be better to use the `estreams.cpp` conversion mechanism here, giving a string representation for each enum value.

CelticMinstrel:
Sorry for the false alarm, I was going through it one commit at a time. I think the only thing missing then is that as far as I know there's no documentation that Boost.Algorithm is now used.

CelticMinstrel:
I pulled this and verified that it builds on Xcode 12, so I think there's no major issues. (I'd like it to be able to build on Xcode 3, mind you, but I'm not able to verify that right now.)

src/game/boe.main.cpp
NQNStudios: It didn't need to be added to the setup of the CI for any of the platforms--it must be a dependency of one of the packages we already install. 

The readme says

> - [Boost](http://www.boost.org/) - Filesystem and System, plus several header-only
>  libraries; if you're picky, you can run scons and see it enumerate exactly which
>  libraries are needed

So I think it could fall under that umbrella.

src/tools/replay.cpp
NQNStudios: It's intended that you call has_next_action() before popping, so returning Element& makes sense

src/game/boe.main.cpp
CelticMinstrel: Nice to see that the readme is contradicting itself slightly – the vcpkg instructions _do_ enumerate every necessary Boost library, after all. (Though I'm not confident there are no others missing.)

src/tools/replay.cpp
```diff
@@ -86,6 +86,9 @@ bool has_next_action() {
 }
 
 std::string next_action_type() {
+	if(next_action == nullptr){
+		throw "Replay error! No action left to check type";
```
CelticMinstrel, line 90: Please throw a `std::string` instead of a `const char*`, since the last-chance catch-all in `main` only handles the former.

That applies both here and in `pop_next_action()`.

CelticMinstrel:
Another thing I was thinking which is probably better for another PR is that we should probably use a proper command line processing library. Given our existing dependencies, the most logical choices in my mind are:

* Clara, which is bundled with Catch2 as a required dependency. Using this would mean that Catch is technically needed to build the game in addition to just the tests, though it wouldn't actually use any of the testing mechanisms of Catch.
* Boost.ProgramOptions, since we're using Boost anyway

That said, I would be open to other possibilities as well. It appears you've chosen to use a subcommand setup (similar to git), and I'm not sure if either of the above libraries can handle that sort of thing. (Boost is more likely to be able to, but I haven't looked.)

CelticMinstrel:
Okay, I think I'm going to go ahead and merge this, but you certainly should add Boost.Algorithm to the README at some point (in the vcpkg install line, maybe also in the place that currently can't be bothered to enumerate the dependencies).

PR #380, opened 2024-07-30
Mac prefs

NQNStudios:
Part 2 of the replay system work/with a focus on getting MacOS replay functionality caught up.

* On Mac, when passing command-line arguments, the apple events function is still called, and it tries to load "replay" and "replayFile.xml" both as exg files, throwing an error message. I fixed that
* I dug into the scary Objective-C++ for mac preferences and wrote some rough code to record the relevant BoE preferences to the action logs and load it back out.
    * I have not tested whether a Mac can replay a log from Windows and vice-versa, but I tried to make it so that will work.
    * Should Objective-C++ code follow naming and style conventions of the C++ codebase, or the conventions of the existing Objective-C++ it's added to?
* Trying to replay every control_focus event was proving a nightmare when accounting for controls that are within containers. I also think it was a mistake to record and replay focus events in the first place, because it looks like they're triggered as side-effects of triggerClickEvent() anyway, meaning that in the current state of things the focus events might be replaying twice. So I reverted adding the `control_focus` as a logged event.
* made led groups work with click replays
    * I've recorded and replayed a session where the game launches, preferences are opened, and window size changes from fullscreen to windowed mode. pretty cool. (That's not included in the PR)

### DIFF
--- a/src/tools/prefs.mac.mm
+++ b/src/tools/prefs.mac.mm
@@ -49,7 +49,7 @@ static NSString* convertKey(std::string keypath) {
 	return key;
 }
 
-NSUserDefaults* getCurrentDefaults() {
+static NSUserDefaults* getCurrentDefaults() {
 	if(replaying){
 		return replayUserDefaults;
 	}else{
--- a/src/tools/prefs.mac.mm
+++ b/src/tools/prefs.mac.mm
@@ -11,6 +11,8 @@
 #include <unordered_map>
 #include "replay.hpp"
 #include <sstream>
+#include <boost/lexical_cast.hpp>
+#include "ticpp.h"
 
 //static const CFStringRef prefsID = CFSTR("com.spidweb.bladesofexile");
 typedef NS_ENUM(NSInteger) {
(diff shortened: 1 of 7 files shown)


src/tools/prefs.mac.mm
```diff
@@ -70,6 +98,58 @@ void clear_pref(std::string keypath) {
...
 bool sync_prefs() {
+	if(recording){
+		std::ostringstream prefs_recording;
+		NSUserDefaults* standard = [NSUserDefaults standardUserDefaults];
+		for(id pref in prefsToRecord) {
+			id object = [standard objectForKey: pref];
+			if(object != nil){
+				prefs_recording << [pref UTF8String] << " = ";
+				NSInteger type = [prefsToRecord[pref] integerValue];
+				bool bvalue;
```
CelticMinstrel, line 175: I think I'd prefer to declare these in the switch statement with the following idiom:

```cpp
switch(int(type)) {
  case kBool: {
    bool bvalue
    // ...
  } break;
}
```

src/tools/prefs.mac.mm
```diff
@@ -94,11 +106,65 @@ void append_iarray_pref(std::string keypath, int value) {
...
+	std::string line;
+	while(std::getline(istream, line)) {
+		if(!istream) {
+			perror("Error reading preferences");
+			return false;
+		}
+		if(line[0] == '#') continue;
+		int eq = line.find_first_of('=');
+		if(eq == std::string::npos) {
+			printf("Error reading preferences: line is not a comment and lacks an = sign:\n\t%s\n", line.c_str());
```
CelticMinstrel, line 124: Is there some reason why we can't use streaming like the rest of the code?

CelticMinstrel:
> Should Objective-C++ code follow naming and style conventions of the C++ codebase, or the conventions of the existing Objective-C++ it's added to?

I think I'd prefer to use the same naming and style conventions across the code. Things will probably never be completely consistent because it's an old and fairly large codebase, but we can at least try to make all the new code be consistent.

src/tools/prefs.mac.mm
NQNStudios: I started by copying over the implementation of load_prefs() for windows/linux:

https://github.com/calref/cboe/blob/10ad8dc9da4a73eb4ad2895b3d1dcc104c48f68a/src/tools/prefs.win.cpp#L105

I wouldn't want to change too much from something we're pretty sure works fine.

CelticMinstrel: Well, I guess we can leave it as is for now; I'd probably want the Linux version to also use streaming there but keeping them as much the same as possible isn't a bad goal.

CelticMinstrel:
Okay, so there's no other reason not to merge this, right? I'll probably merge it tonight then.

NQNStudios:
I think it's ready.

Issue #148, opened 2018-02-15
At startup screen - make party, quit, select "save first" - it doesn't quit


daerogami:
---
This is an item logged from the To-Do and was originally reported by (@CelticMinstrel?). If you are familiar with or can recreate this bug, please provide details here.

NQNStudios:
@CelticMinstrel this bug happens because the logic of this statement (and its repetitions elsewhere) is inverted:
https://github.com/calref/cboe/blob/0a3aa97b59627a05144f9d2b58ea21df003ae42b/src/game/boe.main.cpp#L612

It's not quitting, and it's not saving, either. We just need to delete the `!`, because i think this statement was meant to short-circuit the quitting if they cancel the file browser dialog.

In PR #381 the problematic statement is de-duplicated so it will be best to fix this after #381 is merged.

Incidentally I debugged this one using the replay system, which I've made a separate repo for storing a replay test suite in: https://github.com/NQNStudios/cboe-replay-testing

NQNStudios:
Actually, removing the `!` doesn't fully make this case flow completely correctly. After saving, now it still warns the player "If you quit now, your game will be lost" and that's wrong.

CelticMinstrel:
> After saving, now it still warns the player "If you quit now, your game will be lost" and that's wrong.

This should happen when the mode is anything other than town, outdoors, or startup. It certainly shouldn't ever happen after actually _asking_ to save.

PR #381, opened 2024-08-01
Replay System part 3

NQNStudios:
2 core changes:
* I moved the replay_next_action() call into the actual game loop and only process one action per frame, so things actually redraw in between actions.
* Fixed replaying control clicks inside subclasses of cContainer (so cStack and cScrollPane), which had the same problem that led groups previously had

Aside from that, added recording and replay for several relatively trivial actions.

### DIFF
--- a/src/game/boe.main.cpp
+++ b/src/game/boe.main.cpp
@@ -132,6 +132,7 @@ void handle_missile(bool& need_redraw, bool& need_reprint);
 void handle_get_items(bool& did_something, bool& need_redraw, bool& need_reprint);
 void handle_drop_item(short item_hit, bool& need_redraw);
 void handle_drop_item(location destination, bool& need_redraw);
+void handle_give_item(short item_hit, bool& did_something, bool& need_redraw);
 
 #ifdef __APPLE__
 eMenuChoice menuChoice=eMenuChoice::MENU_CHOICE_NONE;
@@ -315,6 +316,9 @@ void replay_next_action() {
 	}else if(t == "handle_drop_item_location"){
 		location destination = location_from_action(next_action);
 		handle_drop_item(destination, need_redraw);
+	}else if(t == "handle_give_item"){
+		short item_hit = short_from_action(next_action);
+		handle_give_item(item_hit, did_something, need_redraw);
 	}
 
 	advance_time(did_something, need_redraw, need_reprint);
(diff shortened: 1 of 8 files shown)


NQNStudios:
I made the app-quitting logic DRY, with one interesting change to the appleevents.mm file:

A TODO comment in it said that the mac quit event handling was causing a crash. I wasn't actually able to reproduce this crash, so that's weird, but I decided it might be smart while de-duplicating all the quitting logic, to also stop letting apple handle the actual quitting, and let sfml do it for all 3 platforms. So I'm cancelling the apple quit events and queuing fake sfml quit events instead.

Let's see what these CI build errors are...

src/game/boe.main.cpp
```diff
@@ -361,32 +355,36 @@ void handle_events() {
 	cFramerateLimiter fps_limiter;
 
 	while(!All_Done) {
+		if(replaying && has_next_action()){
+			replay_next_action();
+		}else{
```
CelticMinstrel, line 408: I guess it's not super important, but you could've avoided indenting all this code by just adding `continue` on the line above…

src/tools/replay.cpp
```diff
@@ -136,4 +137,18 @@ void decode_file(std::string data, fs::path file) {
...
+}
+
+location location_from_action(Element& action) {
+	location l;
+	std::istringstream sstr(action.GetText());
+	sstr >> l;
+	return l;
+}
+
+short short_from_action(Element& action) {
```
CelticMinstrel, line 149: Seems like a good place to use templates (if you need more than just these two).

src/game/boe.main.cpp
```diff
@@ -117,8 +117,19 @@ extern long anim_ticks;
 static void init_boe(int, char*[]);
 static void showWelcome();
 
+// TODO all these forward declarations of boe.actions.cpp functions might be less
+// than ideal. Maybe they could be moved to boe.actions.hpp and #included here instead
```
CelticMinstrel, line 124: Yes please.

NQNStudios:
I accidentally pushed an unrelated commit from an experiment, the force-push just removed it

CelticMinstrel:
> A TODO comment in it said that the mac quit event handling was causing a crash.

If I recall correctly, the crash occurred when quitting via the menu but not if you quit by some other means (I think possibly even Cmd+Q didn't cause the crash?). It looks like your change probably _does_ fix it, so if there's an open issue for that we should close it when this merges. Though I'll test it first, of course.

NQNStudios:
Whoops, I pushed a commit called Fix #381 when it should have said Fix #148

CelticMinstrel:
You can reword the commit by amending it and force-push it.

PR #385, opened 2024-08-03
Replay system small fix

NQNStudios:


NQNStudios:
I was gonna put more into this PR but I haven't decided what to do next & this one commit is definitely correct and needed so I don't want to hold it up.

CelticMinstrel:
So, it is correct to poll events when replaying?

NQNStudios:
When the replay runs out of actions to replay without quitting the app,
yeah.

CelticMinstrel:
Hmm. I would've expected `replaying` to become false in that case, but alright.

PR #387, opened 2024-08-03
fix #230

### LINKED ISSUE #230: Speed icon shown in party window when speed is not in any abnormal state
As seen on an unrelated screenshot at https://0x0.st/zh0V.png

Original Exile 3 and Blades of Exile did not show the icon when there was no haste/slowness applied.

I personally find those icons distracting/disorienting (icons in party window usually meant some abnormal state that needs attention).

Could be made into a setting maybe?

NQNStudios:
The existence of a 0 in the status map was enough to make the speed icon render. Fixed #230 by checking >0

### DIFF
--- a/src/game/boe.text.cpp
+++ b/src/game/boe.text.cpp
@@ -644,7 +644,8 @@ void draw_pc_effects(short pc) {
 		if(next.first == eStatus::POISON && next.second > 4) placedIcon = 1;
 		else if(next.second > 0) placedIcon = statIcons[next.first].first;
 		else if(next.second < 0) placedIcon = statIcons[next.first].second;
-		else if(next.first == eStatus::HASTE_SLOW) placedIcon = 7;
+		// Uncomment this line to enable the "normal speed" icon from Exile II:
+		// else if(next.first == eStatus::HASTE_SLOW) placedIcon = 7;
 		if(placedIcon >= 0) {
 			rect_draw_some_item(status_gworld, get_stat_effect_rect(placedIcon), pc_stats_gworld, dest_rect, sf::BlendAlpha);
 			dest_rect.offset(13, 0);


CelticMinstrel:
I am not convinced that this issue even exists. The line you changed looks like the line that's supposed to put the "normal speed" icon if you're neither hasted nor slowed, and without the change it looks like it does that, unless 7 is the wrong icon?

NQNStudios:
It hadn't occurred to me that there would *be* a "normal speed" icon but I guess you're right.

staticons

This is the icon spritesheet and counting to 7 I see the one it draws.

So my fix was wrong... but I hate the normal speed icon and think it should be removed

CelticMinstrel:
Are you able to open either of the original versions of the game, downloadable from the Spiderweb website? I theoretically can, but it seems I never copied that over from my old computer so it'll be a pain to get it booted up to try it. I'd like to know whether the "normal speed" icon actually appears in the original game, or if it was something I put back in based on remembering it in Exile 2 and 3 and seeing it on the stat icons sheet.

NQNStudios:
setting up DOSBOX-X/Windows 98 to check. This should come in handy for resolving other questions about how the orriginal worked

NQNStudios:
Confirmed -- there is no Normal Speed icon shown in the official version on windows.

CelticMinstrel:
I wish I could check in Basilisk 2 as well, but… if it wasn't there on Windows, I guess the chances are it also wasn't there in the Mac version. I _will_ probably check at some later point, though.

NQNStudios:
I tried to get Basilisk 2 running on my silicon mac *and* my older model, and it wasn't launching.

CelticMinstrel:
Yeah, it also wasn't launching when I tried it just before posting that. I had it set up and working on the old computer, so if I can get that plugged in, I should be able to check, but that's not as easy as it sounds…

NQNStudios:
No one will miss the "Normal Speed" icon. 😈

CelticMinstrel:
I had the sudden realization that we could just check _official screenshots_. Sure enough, [there's no normal speed icon](http://spidweb.com/blades/BOEshots.html).

NQNStudios:
@CelticMinstrel you can also close #230 now

PR #388, opened 2024-08-03
Fix #327 for windows

### LINKED ISSUE #327: Possible to overwrite a save without confirmation on Linux
(On windows) If I try to save over a file "t.exg" by double-clicking on the existing file "t.exg", I get prompted for confirmation on whether to overwrite.

If I type "t" in the file browser and press enter, it still appends ".exg" to that path and overwrites "t.exg", but it doesn't confirm first.

I don't know what happens on Mac/Linux in this case.

NQNStudios:
This fixes #327 for Windows, and I've seen that on Mac it wasn't a problem. Need to see how Zenity handles this on Linux, though

### DIFF
--- a/src/tools/winutil.win.cpp
+++ b/src/tools/winutil.win.cpp
@@ -198,6 +198,7 @@ static std::string runFileDlog(OPENFILENAMEA& dlg, const std::string& file, bool
 	std::fill(path + file.length(), path + sz, 0);
 
 	dlg.lpstrFile = path;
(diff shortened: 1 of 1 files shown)


NQNStudios:
Linux shares this problem, and zenity doesn't have an obvious option for fixing it easily...

NQNStudios:
So, this will be a lot harder to fix on Linux. The Windows fix should be merged but the issue should actually stay open, labeled for Linux

PR #389, opened 2024-08-03
fix xmllint check in SConscript

NQNStudios:
On my linux installation `which xmllint` returns 0 but the output is empty (because I don't have xmllint). This PR fixes the logic so scons won't keep trying to lint the xml when I do my linux builds


CelticMinstrel:
That's weird – on my machine, `which` exits with 1 if the thing I'm querying doesn't exist. But this is fine, in any case. Though now that I think of it, it might be a good idea to actually mention that it's skipping the lint because it's not installed…

PR #391, opened 2024-08-07
replace rand() and srand(). fix #390

### LINKED ISSUE #390: Replace rand()
https://github.com/calref/cboe/pull/379#discussion_r1694023035

> Seeing this made me realize that we may want to consider moving away from `rand()` in favour of a Mersenne twister… but that would be something for a separate PR.

I was thinking back over this, and remembered a practical/important reason why we should stop using `rand()` for randomness.

The code uses rand() for gameplay randomization, but it also uses rand() to generate unique ids for ui controls. The replay system needs gameplay randomization to be totally deterministic, and so far it hasn't mattered that the UI system is also calling rand(), but if we ever change a dialog xml in a way that the UI system needs to call rand() more or less times, then game replay determinism will be broken between old and new versions of cboe, which defeats the core purpose of replays.

So at the very least we need to use a RNG system where the UI controls' randomized ids are not interfering with the same RNG the gameplay logic uses.

NQNStudios:


src/game/boe.main.cpp
```diff
@@ -343,16 +343,16 @@ void init_boe(int argc, char* argv[]) {
 		Element& srand_element = pop_next_action("srand");
 		
 		std::string ts(srand_element.GetText());
-		srand(atoi(ts.c_str()));
+		game_rand.seed(atoi(ts.c_str()));
```
CelticMinstrel, line 346: This isn't really related to the PR, but I think `std::stoi` might be better than `atoi` here. Also removes the need to call `c_str()`.

src/mathutil.cpp
```diff
@@ -9,6 +9,9 @@
 #include <cstdlib>
 #include "mathutil.hpp"
 
+std::mt19937 game_rand;
+std::mt19937 ui_rand;
```
CelticMinstrel, line 13: I think `ui_rand` can actually be a private static member variable of `cDialog`, since it's not used anywhere else… unless you think it'll be needed in other places?

CelticMinstrel:
I'm a little surprised that this change is so small.

NQNStudios:
> I'm a little surprised that this change is so small.

Luckily most of the rand() calls were wrapped in get_ran()

src/game/boe.main.cpp
NQNStudios: I can do a PR after this one that replaces atoi throughout the codebase, here are all the places we're using it:

```
$ rg atoi
src/location.cpp
245:	l.x = atoi(sstr.str().c_str());
249:	l.y = atoi(sstr.str().c_str());

src/fileio/xml-parser/tinyxml.cpp
578:			*i = atoi( s );
595:			*i = atoi( s->c_str() );
1275:	return atoi (value.c_str ());

src/fileio/xml-parser/tutorial_ticpp.txt
113:int attr = atoi( pszAttr );

src/dialogxml/dialogs/dialog.cpp
544:				eKeyMod mods = static_cast<eKeyMod>(atoi(info["mods"].c_str()));

src/game/boe.main.cpp
267:		eStartButton btn = static_cast<eStartButton>(atoi(next_action.GetText().c_str()));
346:		srand(atoi(ts.c_str()));
```

the xml-parser uses will stay the same. Everywhere else we call atoi() it is a clunky c_str() round-trip conversion like you pointed out

CelticMinstrel: xml-parser is an external library, so yeah, leaving it untouched makes sense. Looks like there aren't that many uses beside the xml-parser. It's likely they were added before 2011 so `stoi` wasn't even available.

CelticMinstrel:
I guess not including `<random>` in `dialog.hpp` would be a tiny argument against my suggestion, but… it doesn't seem like not including it is all that useful of a goal?

PR #395, opened 2024-08-08
Replay scenario list (Fix #392)

### LINKED ISSUE #392: Scenario list is also non-deterministic in replays
Replays track which scenario you open only through the ui events for paging forward/backward in the scenario list and which of the 3 buttons you click on.

So, if you run a replay with a different set of custom scenarios installed, the replay could end up choosing to open the wrong scenario.

This is a very nasty problem, I haven't figured out how to solve it in a way that isn't crazy.

Totally crazy solution:
* When recording the start scenario action, compress the custom scenarios folder into one file and base64-encode it into the replay log.
* When replaying the start scenario action, decode and extract the recorded custom scenarios into the temporary directory, and read the scenario list from there instead of the user's custom scenario folder

NQNStudios:
Fix #392 by implementing the system @CelticMinstrel suggested

### DIFF
--- a/src/game/boe.fileio.cpp
+++ b/src/game/boe.fileio.cpp
@@ -20,6 +20,8 @@
 #include "fileio/fileio.hpp"
 #include "tools/cursors.hpp"
 #include <boost/filesystem.hpp>
+#include "replay.hpp"
+#include <sstream>
 
 #define	DONE_BUTTON_ITEM	1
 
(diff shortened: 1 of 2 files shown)


src/game/boe.fileio.cpp
```diff
@@ -341,28 +341,49 @@ std::vector<scen_header_type> build_scen_headers() {
...
+		std::string scen_file;
+		while(std::getline(in, scen_file)){
 			scen_header_type scen_head;
-			if(load_scenario_header(iter->path(), scen_head))
+			fs::path full_path = scenDir / scen_file;
+			if(load_scenario_header(full_path, scen_head)){
 				scen_headers.push_back(scen_head);
+			}else{
+				scen_header_type missing_scen;
+				missing_scen.intro_pic = missing_scen.rating = missing_scen.difficulty = 0;
```
CelticMinstrel, line 355: Which icon is this actually? The cave one?

PR #396, opened 2024-08-08
record and replay mods on startup buttons. fix #394

### LINKED ISSUE #394: alt+start scenario records incorrectly in replays
I didn't realize the alt-key is significant when clicking startup buttons. need to make that replay action record its state

NQNStudios:
Since there is the one case where a modifier key is relevant in the startup menu, I figure we should handle the modifier keys in the same way dialogs do, because this makes it easier to record and replay that edge case.


CelticMinstrel:
I don't think we want recursive header resolution? I seem to recall intentionally turning that off. Can't you include it as `"tools/keymods.hpp"` instead of using recursive header resolution?

Issue #372, opened 2024-06-21
Still using 100% cpu for a turn-based game that needs 0% cpu.


clort81:
There's no game logic running between turns / moves, at all.  It ran on a 20mhz 16-bit machine, originally.

So, has anyone done a fork that fixes this?

Please link if there is one.

CelticMinstrel:
I would guess that the chief reason for this may be the animated terrain, which is probably CPU-bound. Logically, that could probably be done on the GPU, though. I am not aware of anyone who has tried this or any other changes that could affect CPU usage, however.

NQNStudios:
While recently going through startup code to fix #366, this part of the code flagged my attention:

https://github.com/calref/cboe/blob/5d7332562195d4a5db5547bb6797c143c506848c/src/game/boe.startup.cpp#L131
```
	while(sound_going(95)) {
		draw_splash(pict_to_draw, mainPtr, logo_from);
		handle_splash_events();
	}
```

This code will update/draw as many times per second as the processor and GPU can handle. Which is not good--typically a game will lock its frame rate to a maximum of 60 or 30 by delaying at the end of every frame if time was used than needed.

I don't know if BoE is running as fast as it can at all times, or just at startup, but if it is, that could definitely be the culprit.

NQNStudios:
Ah, this could be limiting the frame rate: https://github.com/calref/cboe/blob/5d7332562195d4a5db5547bb6797c143c506848c/src/game/boe.main.cpp#L340

CelticMinstrel:
> I suspect this _is_ the source of the problem

That loop is only during startup, so I'm fairly sure it can't be the source of a general problem.

NQNStudios:
Yeah.

NQNStudios:
I found another place where event loops are spun up without frame rate limiting:
https://github.com/calref/cboe/blob/c251fee834a2c078b51d852f117e1442fde85771/src/dialogxml/widgets/control.hpp#L326

For example:
https://github.com/calref/cboe/blob/c251fee834a2c078b51d852f117e1442fde85771/src/dialogxml/widgets/control.cpp#L218

Still I don't know if this explains anything.

CelticMinstrel:
That would explain jumping to 100% CPU when you press and hold your mouse on something in a dialog. That isn't what's described here.

NQNStudios:
I'm making a single issue to track CPU-related concerns: #450 

[...]

Issue #321, opened 2023-01-22
Still redrawing screen constantly although nothing changed?


clort81:
Is cboe still redrawing screen although nothing changed on the game screen?

Any turn-based game without continuously running onscreen animations, has 0 cpu consumption unless the player makes a move.  (Such as original Blades of Exile code does.)

This could be nice on battery powered devices.

NQNStudios:
I don't think it is (at least when dialogs are open), which is why #208 is a problem.

Continuously redrawing the game is the only way to completely fix #208 as long as BoE uses floating windows for dialogs, and it's standard for modern games, but maybe if it's implemented it could be a togglable option?

NQNStudios:
I think a strong argument could be made that doing away
with floating windows for dialogs would be a good thing, though. JV's
newer games don't do it, for good reason.

CelticMinstrel:
> (at least when dialogs are open)

As I recall, the main loop doesn't even run while dialogs are open. The dialog has its own separate event loop.

I'm pretty sure both dialogs and the main application both redraw continuously.




> I think a strong argument could be made that doing away
> with floating windows for dialogs would be a good thing, though. JV's
> newer games don't do it, for good reason, from a UX and portability perspective.

You're right that JV stopped doing it eventually, but I would like to maintain the original aesthetic of the game.
If it's easy (or becomes easy at a later date), we could perhaps offer a setting to stuff the dialogs into the main window. I would only offer that in the game though, not in either of the editors. It may not work, however, as I think some of the dialogs (choose spell? training?) are actually larger than the main window.

clort81:
Thanks for the responses!   In case anyone reading doesn't know, some background info:

The bug referenced is at root a legacy problem leftover from 1980s and 1990s workstations, which didn't have enough RAM for the windowing system to buffer all window contents of applications.  When one application got overlapped by another window, then 'exposed', the application had to be notified to redraw itself.

Eventually we can expect all linux users will be using compositing Xorg or Wayland, which handles that problem.

Could you realistically envison someday having a display client (like, cough, a webbrowser) messaging the CBOE engine and say 'Hey, I'm ready to display what's happening, can you tell me what has changed in menus and game world?' - and then receiving the stuff to update on screen?

NQNStudios:
I'm making a single issue to track CPU-related concerns: https://github.com/calref/cboe/issues/450

I do want to answer something from this discussion which never got addressed:

[...]

Issue #227, opened 2020-01-08
dialogxml is too slow/cpu-hungry


x-qq:
Currently, simply having a dialogxml-backed window on screen consumes almost 100% of a powerful CPU core.

![dialogxml_too_slow](https://user-images.githubusercontent.com/55463047/71991069-3746c480-323d-11ea-954f-570f42905021.png)

For reference, this is Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz.

This is not a reasonable CPU load for some static text and pictures, so I think dialogxml requires some optimization.

CelticMinstrel:
Hmm. I can guess that maybe part of the cause is that it's redoing the text layout (wrapping) every frame? (I don't know whether it _is_ doing that though.)

x-qq:
callgrind shows this

![out](https://user-images.githubusercontent.com/55463047/72074240-36c03380-32fa-11ea-9a4b-8db122cb97fa.png)

clort81:
Can we change OpenBoE to only use CPU when user input is received?  Just stop the process and children until mouse or keyboard event goes to it?

CelticMinstrel:
> callgrind shows this

Not quite certain but it seems like that might support my theory, as there seems to be a lot of green around the `win_draw_string` area.

> Can we change OpenBoE to only use CPU when user input is received? Just stop the process and children until mouse or keyboard event goes to it?

I think it might be possible, since the game is turn-based. The only possible complication would be the animations that should continue to work even while time isn't passing.

x-qq:
CPU load ws reduced in 1fcbd8e922c59a8f4a58e4c16a4c5b966a87091d.
But we still need some form of caching, because currently widgets do too much during redraw.

x-qq:
Updated callgrind:

![out svg](https://user-images.githubusercontent.com/55463047/72829328-3d9a5f00-3c87-11ea-89f0-334aa15e86e3.png)

NQNStudios:
I'm making a single issue to track CPU-related concerns: https://github.com/calref/cboe/issues/450

Note that this specific issue has been *mostly* solved. We know that multiline text fields are still slow, something that I've noted in the new issue, but those only exist in the scenario editor which isn't the current top priority.

Issue #238, opened 2020-01-18
Keyboard buffer in combat doesn't clear.


clort81:
Holding down space in combat for a while fills keyboard buffer with 'wait'.  During combat, the keyboard input buffer isn't cleared during next move, so your whole party will wait tens or hundreds of times, and die.

This is possibly related to the diagonal combo-movement bug, where the next 'turn' gets keyboard input from previous turn.

NQNStudios:
> This is possibly related to the diagonal combo-movement bug, where the next 'turn' gets keyboard input from previous turn.

Linking #225 to this thread--it's the issue in question.

NQNStudios:
Can't reproduce. I think it's fixed.

PR #397, opened 2024-08-09
Fix CPU usage

NQNStudios:
We have 3 open issues that reference unreasonably high CPU usage. Using XCode to debug CPU usage, I believe I found all the cases that have been reported and fixed them.

#227 - DialogXMLs with wrapped text are doing all of their processing every time `draw()` is called. So whenever such a dialog is onscreen, CPU usage jumps to about 90% for me until it's closed. I fixed this by only calling `draw()` on frames where an input event occurs. This still will be drawing more than necessary, probably, and a fix could go further by passing a `need_redraw` flag to the event handlers instead of treating any event as a reason to redraw.

#321 and #372 - these issues were more vague about which specific game states were using 100% CPU, and asked for a total overhaul of the game loop. From what I can tell, though, CPU spikes in the intro splash screen, when clicking on controls, and when dialogxmls are wrapping text, are the real problems. Theoretically it would be ideal to convert the whole game to an event-responsive code style where CPU usage stays near 0 in between player turns, but I don't think that's really warranted as long as I've actually found the aberrant cases and fixed them.

To fix the mini event loops that every control has in its `handleClick()` method I added an fps_limiter to the method signature and have them finish their frames with it. This required changing an absurd number of function signatures, and at first I thought just making one `fps_limiter` global would be better, but I realized one advantage of changing all these function signatures is that it becomes more clear in the code when a subcomponent is expecting to spawn its own internal event loop.

### DIFF
--- a/src/dialogxml/dialogs/dialog.cpp
+++ b/src/dialogxml/dialogs/dialog.cpp
@@ -544,7 +544,7 @@ void cDialog::handle_events() {
 			eKeyMod mods = static_cast<eKeyMod>(std::stoi(info["mods"]));
 			controls[info["id"]]->triggerClickHandler(*this, info["id"], mods);
 		}else{
-			while(win.pollEvent(currentEvent)) handle_one_event(currentEvent);
+			while(win.pollEvent(currentEvent)) handle_one_event(currentEvent, fps_limiter);
 		}
 
 		// Ideally, this should be the only draw call that is done in a cycle.
@@ -556,7 +556,7 @@ void cDialog::handle_events() {
 }
 
 // This method handles one event received by the dialog.
-void cDialog::handle_one_event(const sf::Event& currentEvent) {
+void cDialog::handle_one_event(const sf::Event& currentEvent, cFramerateLimiter& fps_limiter) {
 	using Key = sf::Keyboard::Key;
 
 	cKey key;
@@ -671,7 +671,7 @@ void cDialog::handle_one_event(const sf::Event& currentEvent) {
 		case sf::Event::MouseButtonPressed:
 			key.mod = current_key_mod();
 			where = {(int)(currentEvent.mouseButton.x / ui_scale()), (int)(currentEvent.mouseButton.y / ui_scale())};
-			process_click(where, key.mod);
+			process_click(where, key.mod, fps_limiter);
 			break;
 		default: // To silence warning of unhandled enum values
 			break;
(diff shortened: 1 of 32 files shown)


NQNStudios:
I also checked having animated terrains on the screen and the CPU usage didn't spike for those.

NQNStudios:
- [ ] Walking on lava spikes CPU to 100%

NQNStudios:
- [ ] starting outdoor combat spikes CPU

CelticMinstrel:
> Theoretically it would be ideal to convert the whole game to an event-responsive code style where CPU usage stays near 0 in between player turns,

I don't think animated terrain would work with this? But if there's a way to have both, then sure.

NQNStudios:
force-pushed to rebase.

NQNStudios:
I decided instead of only redrawing dialog when an event happens, to actually delve into the multiline string wrapping that causes the CPU spikes. Because it is still pretty bad if I can get a 60% CPU spike by fiddling with the mouse when multiline text is visible.

CelticMinstrel:
What was holding this up?

NQNStudios:
I got confused about the `event_sleep()` feature I was adding, whether it made theoretical sense or not, as a solution to any  problems. Which I think it might, maybe not for CPU improvement, but because #238 might be caused by input events not being polled and flushed while animations are running `sf::sleep()`.

When I get too confused or frustrated I usually switch to other tasks for a while.

NQNStudios:
Also caching the line wrapping was making multiline input fields behave VERY wildly, which doesn't feel like a very good priority to fix because I can't think of any multiline input fields in the game. Aren't those all in the scenario editor?

NQNStudios:
I don't have a good way set up to profile the code execution. With Xcode I can profile CPU usage over time but I have no way to delve into the spikes and see which code was running at the time. Do you know of a way?

CelticMinstrel:
> Also caching the line wrapping was making multiline input fields behave VERY wildly, which doesn't feel like a very good priority to fix because I can't think of any multiline input fields in the game. Aren't those all in the scenario editor?

Yeah, there are no multiline text fields in the game itself, only in the scenario editor. Pretty sure even the character editor doesn't have them.

> Do you know of a way?

I'm not very familiar with Xcode profiling, but I thought it had a way to tell you what function was running during a spike…

NQNStudios:
> I'm not very familiar with Xcode profiling, but I thought it had a way to tell you what function was running during a spike…

Maybe that's possible if you *build* through Xcode--a lot of debugging features are limited for me because I'm building through scons. Like, I can't browse through the source files to set breakpoints. It's really annoying.

I haven't been able to get the xcode build to work for me locally.

CelticMinstrel:
Hmm… I bet there's a way to set up something like a "source mapping" to map the symbol information actually in the file to the source files that Xcode is actually looking at – a scons build would be built from the build directory, so the input source files are actually different, I think. But I'm not sure how you would do that in either lldb or an Xcode scheme…

NQNStudios:
Every change in this PR related to CPU has been merged through other ways. I've taken note of the rest of the potentially useful commits, which I will reincorporate in an appropriately-themed future PR.

PR #403, opened 2024-08-10
WIP Command-line reform

CelticMinstrel:
This makes use of the Clara command-line library included with Catch to do proper parsing of command-line arguments.

This fixes an issue where the scenario editor complains about a missing file when launched from Xcode, because Xcode passes an unrecognized argument which is interpreted as a filename. The Clara parser seems to ignore unrecognized arguments, so using it makes this work. It does unfortunately require using flag syntax instead of subcommand style for the replay system.

If this looks mostly okay, it might be better to update to the Lyra library instead, which is similar in substance but is still being maintained. Unlike Clara, Lyra has support for subcommands, so we wouldn't need to change the existing command-line syntax. I don't know if it ignores unrecognized arguments like Lyra does.

Alternatively, we could consider using Boost.ProgramOptions for command-line parser. This is a _much_ bulkier library, however.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -352,9 +352,12 @@ if not env.GetOption('clean'):
 	check_lib('sfml-audio', 'SFML-audio')
 	check_lib('sfml-graphics', 'SFML-graphics')
 
-	# If building the tests, make sure Catch2 is cloned
-	if 'test' in targets and not path.exists('deps/Catch2/README.md'):
+	# Make sure Catch2 is cloned
+	if not path.exists('deps/Catch2/README.md'):
 		subprocess.call(["git", "submodule", "update", "--init", "deps/Catch2"])
+	
+	# We also use the Clara bundled with Catch in the main program
+	env.Append(CPPPATH=[path.join(os.getcwd(), 'deps/Catch2/include/external')])
 	
 	# Make sure cppcodec is cloned
 	if not path.exists('deps/cppcodec/README.md'):
--- a/proj/vs2017/Blades of Exile.vcxproj
+++ b/proj/vs2017/Blades of Exile.vcxproj
@@ -78,7 +78,7 @@
       <WarningLevel>Level3</WarningLevel>
       <DebugInformationFormat>EditAndContinue</DebugInformationFormat>
       <Optimization>Disabled</Optimization>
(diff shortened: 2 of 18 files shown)


CelticMinstrel:
@NQNStudios Thoughts?

NQNStudios:
My command-line arg code is kind of a mess. It can definitely be improved with a good library on the back-end. If Clara is a part of Catch, that seems like a great option.

Flags are probably better than subcommands, because when you mis-type a subcommand currently, boe tries to open it as a file, which is never correct. Starting the subcommands with a `-` might help disambiguate that problem

NQNStudios:
Be careful for how the apple events snatch up command-line arguments which will then also be processed by our own arg processing. That tripped me up at least once

CelticMinstrel:
> Clara is a part of Catch, that seems like a great option.

It is part of Catch, but it is archived and no longer maintained. I would assume Catch v3 is no longer using it, but as we require C++11 support we can't upgrade past v2 anyway.

NQNStudios:
I mentioned the apple events weirdness [here](https://github.com/calref/cboe/pull/380#issue-2436760708):

> * On Mac, when passing command-line arguments, the apple events function is still called, and it tries to load "replay" and "replayFile.xml" both as exg files, throwing an error message. I fixed that

and addressed it with this change:

For some reason the apple event that's set up to handle when a file gets dragged onto the app icon, also handles direct command line argument. So when I launched with "replay file.xml" it was trying to load a party called "replay"
https://github.com/calref/cboe/pull/380/commits/05111721616abc23726d176de18c4e5862e1692e

CelticMinstrel:
You're saying that MacOS sends an open file event to the application when it finds a command line argument? That doesn't even make sense… I don't get it…

NQNStudios:
It seemed pretty outrageous to me too, but I just patched around it and moved on. 😢

NQNStudios:
I checked this branch out, built, and made a quick recording. Parts of the replay are visually messed up, like when the splash screen ends I see the startup button that gets clicked drawn on top of it, but the menu doesn't get a chance to draw itself in full first. But when my party gets into a scenario and I'm replaying with `--replay-speed 1`, it does indeed seem like they move around at a rate  of 1 turn per second as expected.

So I'm thinking that what you've done *does* work, but certain things the replay system replays, happen outside of `replay_next_action()` in `handle_events()`. For example, `dialog.cpp` is handling replay actions within its event loop, and that has its own framerate limiter that doesn't pick up the CLI arg.

If we do change animations to not have their own event loops, then this way of specifying replay FPS is gonna mess those up and make them last forever.

CelticMinstrel:
> then this way of specifying replay FPS is gonna mess those up and make them last forever.

Do you have a better way of specifying replay FPS?

NQNStudios:
Still thinking about this. The cleanest way I've come up with so far, would be to put an fps_limiter in replay.cpp and have `pop_next_action()` end its frame before returning. While this would seem to be an unexpected side-effect for a function called `pop_next_action()`, it would ensure that every event loop that replays anything, stops first to let the user see what happened in the last frame. While still leaving the main FPS system alone.

CelticMinstrel:
Okay, I'll try that.

NQNStudios:
The start menu still probably won't draw with this, because the issue there I'm guessing is that the splash screen ends in the same frame the button action gets picked up, so sleeping in pop_next_action() would just draw the splash screen longer.

Issue #47, opened 2015-09-20
Text edit boxes are laggy


CelticMinstrel:
I think this is related to text rendering, but not quite sure. It seems to be worse on Windows, but does affect OSX as well.

CelticMinstrel:
Considering using the TGUI text box to fix this, but I'll need to a) evaluate which features of the current box would still need to be implemented on top of it and b) whether I can get TGUI working with my personal build processes.

NQNStudios:
@CelticMinstrel Did you observe this mainly in the scenario editor, as the strings you were typing got longer and longer?

NQNStudios:
I see that in some scenario editor windows there are multi-line messages in addition to text fields, so those text fields would be expected to be laggy even when empty, because of lag induced by wrapping the messages every frame

CelticMinstrel:
I think I've _only_ observed this in the scenario editor. I don't remember if it has been improved since I first noticed it, but I remember it taking seconds to move the insertion point in some fields.

NQNStudios:
I refactored them to try to use `calculate_line_wrapping()` and it seems to break mouse input for moving the cursor and dragging to select text....

PR #406, opened 2024-08-13
Fix cpu rebase

NQNStudios:
This rebases the first set of changes in #397:

* Get CPU usage down from 100% during splash screens and control clicks
* Make text line-wrapping cacheable, and cache the wrapping for cTextMsg

A couple of things to do before merging:

- [x] Make cTextMsg re-line-wrap when the text changes
~~- [ ] Make other callers of `win_draw_string()` cache their line breaks, where applicable~~ (in a later PR)

As far as I know, this would solve the only CPU-hogging component of the DialogXML system, thus fixing #227


NQNStudios:
Applying this fix to `cField` could potentially fix or at least improve #47 as well?

NQNStudios:
> I refactored them to try to use calculate_line_wrapping() and it seems to break mouse input for moving the cursor and dragging to select text....

In light of that, I want to call this PR ready (if the build succeeds) and work on putting cTextField optimization in a separate one.

PR #411, opened 2024-08-23
fix #405 and fix #225

### LINKED ISSUE #225: Moving one space diagonally sometimes moves character two spaces.
In combat, i try to move a character one space diagonally down-left by quickly hitting down-left.

In some situations, it moves the character two diagonal spaces.  This seems to be a bug to me (cannot reproduce in original game).

In another instance, hitting up-left to attack an enemy diagonally resulted in my next character moving left (into a force field and taking damage).

Somehow the 'diagonal combos' seem to be not clearing input buffer after one 'get'.

### LINKED ISSUE #405: Arrow keys movement is unpredictable on Silicon macOS 14.5
Something truly bizarre is happening with BoE keyboard input on my newer Mac. I can't tell if it's affecting letter hotkeys (g for get works fine) but arrow key movement is definitely very broken.

Of interest is this console output:
> 0x10000088c: TCC deny IOHIDDeviceOpen
> 0x10000088c: TCC deny IOHIDDeviceOpen
> We got a keyboard without any keys.

I don't know if the first 2 lines are related to the 3rd one, but the 3rd one comes from sfml and I think it's because newer MacOS requires apps to ask for permission to listen to the keyboard: https://en.sfml-dev.org/forums/index.php?topic=27734.0

Maybe I accidentally didn't give BoE this keyboard permission the first time I launched it.

Anyway, the actual problem is that most of the arrow keys do nothing, and the ones that do something, will actually *change which direction they control* after repeated use.

NQNStudios:
Partly implemented fix for arrow key handling on mac, which also introduces a frame buffer for arrow key combinations that would solve #225.

The problem now is that diagonals formed by two arrow keys, don't repeat properly. So I'm trying to disable key repeating when diagonals are used, and turn it back on in other cases. Which is just messy and not good. Needs more tinkering.

### DIFF
--- a/src/game/boe.main.cpp
+++ b/src/game/boe.main.cpp
@@ -387,19 +387,17 @@ using Key = sf::Keyboard::Key;
 std::map<Key,int> delayed_keys;
 const int ARROW_SIMUL_FRAMES = 3;
 
-// Terrain map coordinates to simulate a click for 8-directional movement/waiting
-const int SOUTHWEST = 1;
-const int SOUTH = 2;
-const int SOUTHEAST = 3;
-const int WEST = 4;
-const int EAST = 6;
-const int NORTHWEST = 7;
-const int NORTH = 8;
-const int NORTHEAST = 9;
+// Terrain map coordinates to simulate a click for 8-directional movement/waiting,
+// ordered to correspond with eDirection
 location terrain_click[10] = {
-	{150,185},{120,215},{150,215},{180,215},
-	{120,185},{150,185},{180,185},
-	{120,155},{150,155},{180,135}
+	{150,155}, // north
+	{180,135}, // northeast
+	{180,185}, // east
+	{180,215}, // southeast
+	{150,215}, // south
+	{120,215}, // southwest
+	{120,185}, // west
+	{120,155}, // northwest
 };
 
 void fire_delayed_key(Key code) {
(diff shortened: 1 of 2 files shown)


src/game/boe.main.cpp
```diff
@@ -387,6 +387,21 @@ using Key = sf::Keyboard::Key;
 std::map<Key,int> delayed_keys;
 const int ARROW_SIMUL_FRAMES = 3;
 
+// Terrain map coordinates to simulate a click for 8-directional movement/waiting
+const int SOUTHWEST = 1;
```
CelticMinstrel, line 391: Pretty sure we already have these constants in an enum somewhere? Probably `eDirection`?

NQNStudios: oh, eDirection has

```
	DIR_N = 0,
	DIR_NE = 1,
	DIR_E = 2,
	DIR_SE = 3,
	DIR_S = 4,
	DIR_SW = 5,
	DIR_W = 6,
	DIR_NW = 7,
	DIR_HERE = 8,
```

But these don't correspond to indices of the terrain_click array.

CelticMinstrel: Is there a reason the terrain_click array has to be ordered that way? (I'm genuinely not sure, for the record.)

NQNStudios: Good question. I believe it's because of this:

https://github.com/calref/cboe/blob/336d0e626b23b50f6193a178ae3593b475cbf952/src/game/boe.actions.cpp#L1639-L1647

The numpad keys are in an array in ascending order,  and terrain_click is ordered to correspond with them.

Since I've copy/pasted the terrain_click array into `fire_delayed_key()` that instance of it could be reordered to correspond with `eDirection` values and remove the constants I made.

CelticMinstrel: I see, seems the intent was likely for `keypad[chr - '0']` to produce the keypad key for a given digit, and `terrain_click` is required to be sorted in the same order.

It's a bit ugly to have two copies of `terrain_click` with two different orders, but I guess that would work. I do suggest at least adding a TODO comment though.

src/game/boe.main.cpp
```diff
@@ -387,19 +387,17 @@ using Key = sf::Keyboard::Key;
...
-	{120,185},{150,185},{180,185},
-	{120,155},{150,155},{180,135}
+	{150,155}, // north
+	{180,135}, // northeast
+	{180,185}, // east
+	{180,215}, // southeast
+	{150,215}, // south
+	{120,215}, // southwest
+	{120,185}, // west
+	{120,155}, // northwest
```
CelticMinstrel, line 400: I was momentarily confused by the fact that there are now 8 entries instead of 10, but looking at the code below, it makes sense – the other version has 2 identical entries for waiting (numpad 0 or 5), but this one doesn't handle waiting at all.

But in that case, I think the comment and the array size should be changed.

PR #412, opened 2024-08-24
Record more

NQNStudios:
Starting to record and replay more things, mostly menu items and help buttons.

### DIFF
--- a/src/game/boe.actions.cpp
+++ b/src/game/boe.actions.cpp
@@ -3265,3 +3265,13 @@ bool check_for_interrupt(){
 	}
 	return false;
 }
+
+void handle_new_pc_graphic() {
+	if(recording){
+		record_action("handle_new_pc_graphic", "");
+	}
+	short choice = char_select_pc(1,"New graphic for who?");
+	if(choice < 6)
+		pick_pc_graphic(choice,1,nullptr);
+	draw_terrain();
+}
--- a/src/game/boe.actions.cpp
+++ b/src/game/boe.actions.cpp
@@ -3275,3 +3275,14 @@ void handle_new_pc_graphic() {
 		pick_pc_graphic(choice,1,nullptr);
 	draw_terrain();
 }
+
+void handle_rename_pc() {
+	if(recording){
+		record_action("handle_rename_pc", "");
+	}
+	short choice = char_select_pc(1,"Rename who?");
+	if(choice < 6)
+		pick_pc_name(choice,nullptr);
+	put_pc_screen();
+	put_item_screen(stat_window);
+}
--- a/src/game/boe.actions.cpp
+++ b/src/game/boe.actions.cpp
@@ -237,6 +237,14 @@ void init_screen_locs() {
 	pc_help_button.right = 267;
 }
 
+// Some actions directly show a dialog. This handles that, and records it.
+void show_dialog_action(std::string xml_file) {
+	if(recording){
+		record_action("show_dialog_action", xml_file);
+	}
+	cChoiceDlog(xml_file).show();
+}
+
 bool prime_time() {
 	return overall_mode == MODE_OUTDOORS || overall_mode == MODE_TOWN || overall_mode == MODE_COMBAT;
 }
@@ -1343,7 +1351,7 @@ bool handle_action(const sf::Event& event, cFramerateLimiter& fps_limiter) {
 			rectangle help_button = pc_help_button;
 			help_button.offset(pc_win_ul);
 			arrow_button_click(help_button);
-			cChoiceDlog("help-party").show();
+			show_dialog_action("help-party");
 		}
 		for(int i = 0; i < 6; i++)
 			for(auto j : pc_buttons[i].keys())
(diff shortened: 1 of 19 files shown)


src/game/boe.actions.cpp
```diff
@@ -2627,21 +2627,24 @@ void switch_pc(short which) {
...
 }
 
-void drop_pc(short which) {
-	std::string choice;
-	
-	choice = cChoiceDlog("delete-pc-confirm",{"yes","no"}).show();
-	if(choice == "no") {
-		add_string_to_buf("Delete PC: Cancelled.");
-		return;
+void drop_pc() {
```
CelticMinstrel, line 2630: I kinda think it makes more sense to keep the function taking one parameter… or maybe rename it?

NQNStudios: If it's kept keeping only one parameter, then the choice of PC from 1-6 will have to be done outside of `drop_pc()` at every callsite, which is icky. We're only calling `drop_pc()` from one place, but choosing which PC to drop is an inherent part of the `drop_pc()` process, so why not put it in the function?

CelticMinstrel: Well, to me the name "drop_pc" implies it should take a parameter…

NQNStudios: maybe `handle_drop_pc()`?

src/game/boe.actions.cpp
```diff
@@ -2280,6 +2280,21 @@ void do_save(short mode) {
 	redraw_screen(REFRESH_TEXT);
 }
 
+void do_abort() {
+	if(recording){
+		record_action("do_abort", "");
+	}
+	if(overall_mode != MODE_STARTUP) {
```
CelticMinstrel, line 2287: Isn't this wrong? Shouldn't it be checking `party_in_memory`? If you have a party open at the startup screen and choose abort, I think it should confirm, right?

Though I guess this is exactly what it does right now, so that would be a separate issue…

src/tools/replay.cpp
```diff
@@ -89,6 +93,32 @@ void record_action(std::string action_type, std::map<std::string,std::string> in
 	log_document.SaveFile(log_file);
 }
 
+void record_field_input(cKey key) {
+	std::map<std::string,std::string> info;
+	std::ostringstream sstr;
+
+	sstr << key.spec;
```
CelticMinstrel, line 100:
```suggestion
	sstr << std::boolalpha << key.spec;
```

src/tools/replay.cpp
```diff
@@ -162,3 +192,29 @@ short short_from_action(Element& action) {
...
 	return s;
 }
+
+cKey key_from_action(Element& action) {
+	auto info = info_from_action(action);
+	cKey key;
+	int enum_v;
+
+	std::istringstream sstr(info["spec"]);
+	sstr >> key.spec;
```
CelticMinstrel, line 202:
```suggestion
	sstr >> std::boolalpha >> key.spec;
```

src/game/boe.actions.cpp
CelticMinstrel: Sure.

src/game/boe.actions.cpp
NQNStudios: I agree the condition of the if should be `party_in_memory`, not what it is.

Other places where quitting, there's no check to skip confirming if a save just happened. So I don't know how to easily implement that.

But changing the `if` condition, is trivial so I'm happy to put that in this PR.

NQNStudios:
This covers a lot of tedious actions, I'm ready to be done with it :P

src/game/boe.main.cpp
```diff
@@ -360,6 +360,19 @@ static void replay_next_action() {
...
+		int enum_v;
+		sstr >> enum_v;
+
+		eSkill mode = static_cast<eSkill>(enum_v);
+
+		short force_spell;
+		sstr.str(info["force_spell"]);
+		sstr >> force_spell;
+
+		display_spells(mode, force_spell, 0);
```
CelticMinstrel, line 377: Since you're changing this anyway, maybe fix the use of 0 as `nullptr` while you're at it?

src/game/boe.main.cpp
```diff
@@ -373,6 +373,12 @@ static void replay_next_action() {
 		sstr >> force_spell;
 
 		display_spells(mode, force_spell, 0);
+	}else if(t == "display_skills"){
+		std::istringstream sstr(next_action.GetText());
+		sstr >> enum_v;
+
+		eSkill force_skill = static_cast<eSkill>(enum_v);
+		display_skills(force_skill, 0);
```
CelticMinstrel, line 383: Same here.

src/game/boe.actions.cpp
```diff
@@ -3310,4 +3310,15 @@ void handle_rename_pc() {
 		pick_pc_name(choice,nullptr);
 	put_pc_screen();
 	put_item_screen(stat_window);
+}
+
+void handle_menu_automap() {
```
CelticMinstrel, line 3315: Is there a reason for this to be different from showing the automap via hotkey or button click?

NQNStudios: A good reason? Not sure. But as it is, through the menu, you get a buffer message if it doesn't open. Through the button, if it doesn't open, there's no message.

CelticMinstrel: That's… kinda weird. If it doesn't open for whatever reason, surely it should tell you why?

NQNStudios: I could make that change.

Issue #409, opened 2024-08-23
Record and replay the remaining actions


NQNStudios:
Known actions/systems that aren't supported yet in the replay system:
- [x] Look/search
- [x] buying items
- [x] selling items
- [x] start conversation with NPC
- [x] Ask NPC through preset buttons
- [x] Ask NPC through dynamic (word) buttons
- [x] Rest
- [x] Cast spells
- [x] Debug actions
- [x] z key to show inventory
- [x] type ? for contextual help (some cases are already handled by show_dialog_action)
- [x] use space
- [x] quests
- [x] special items
- [x] interrupt
- [x] rest of the PC pane
- [x] more ? buttons in different modes?
- [x] tab in dialogs with multiple text fields
- [x] changing a text field's selection with the mouse
- [x] delete key in text fields
- [x] shift+movement key selection in text fields
- [x] show item info
- [x] scroll up/down in item list

There will be more that I haven't thought of yet.

NQNStudios:
- [ ] canceling a save dialog

CelticMinstrel:
I was making a great big list of everything, but I was composing it on GitHub and lost it when the browser suddenly decided to switch pages.

I'll give a general overview and a few examples of what's missing though:

* Menu actions. Pretty much all the menu actions are missing. There are a few vital actions squirrelled away in the menus.
* Debug actions. It's debatable whether these are needed, but they exist. Press Shift+D and / for a nearly-complete list (it's missing Shift+I).
* Some keyboard-only actions, like Z, B, Shift+L, Shift+/
* Jobs and Special Items (in the inventory pane) - there exists no scenario that uses jobs, so that one might be a little tough to test
* Nearly all the PC pane actions are missing – almost everything in that panel is clickable
* Help buttons - pretty much every area has a little help button
* Use Space - closing a door, for example

I'm not sure why you listed picking up items as unsupported – it works right now.

NQNStudios:
I forgot getting items uses a dialog, which made it trivial to record and replay so I already did it.

CelticMinstrel:
Aside from talking and shopping, I think the main ones you still haven't covered in #412 are things in the party and inventory panels.

* Switch PCs
* PC info
* Check PC HP
* Check PC SP
* Item info
* Special Items (with use and info buttons)
* Quests (I think these just have an info button, but I'm not sure)
* Does give item work now in #412? I saw you handled text input but not sure if that covers give.

And probably a few toolbar buttons:

[...]

PR #416, opened 2024-08-27
Record more actions

NQNStudios:
- I saw another CPU-hungry event loop to slow down.
- record/replay looking at things
- record/replay moving the camera while in look mode

Linking to #409 for convenience


NQNStudios:
@CelticMinstrel My wifi went out, I tried reconnecting to Liberachat from my phone's hotspot, and I accidentally got banned for spamming as I clicked reconnect too many times 😬

CelticMinstrel:
I mentioned that maybe your confusing thing in spellcasting was because the failbit had been set on the stream. But that implies that reading in `which_type` failed somehow, so I'm not sure if it fits. If that _is_ the issue, calling `.clear()` would make it work.

src/game/boe.actions.cpp
```diff
@@ -780,7 +780,12 @@ static void handle_use_space(location destination, bool& did_something, bool& ne
 	put_item_screen(stat_window);
 }
 
-static void handle_bash_pick_select(bool& need_reprint, bool isBash) {
+void handle_bash_pick_select(bool& need_reprint, bool isBash) {
+	if(recording){
+		std::ostringstream sstr;
+		sstr << std::boolalpha << isBash;
+		record_action("handle_bash_pick_select", sstr.str());
```
CelticMinstrel, line 787: Instead of a boolean, would it maybe be better for this to be two different actions?

NQNStudios: I was thinking the same thing while I wrote this code. I'll do that.

NQNStudios:
I've recorded and replayed every debug action except for 2 of them, which seem almost the exact same, so I think they should be combined using `BOOST_FALLTHROUGH`:

https://github.com/calref/cboe/blob/54050a7cbbe4167e86d3c1149a636180f8d39c93/src/game/boe.actions.cpp#L1828-L1846

https://github.com/calref/cboe/blob/54050a7cbbe4167e86d3c1149a636180f8d39c93/src/game/boe.actions.cpp#L1907-L1920

NQNStudios:
I found the source of issue #418, and it made sense to apply the fix as part of this PR because I've refactored the code in question.

NQNStudios:
All done! Switching all the stringstream stuff to boost::lexical_cast actually fixed at least one bug.

This should close #409 unless we think of anything else.

src/game/boe.actions.cpp
```diff
@@ -495,18 +485,9 @@ void handle_pause(bool& did_something, bool& need_redraw) {
...
-		sstr.str("");
-		sstr << std::boolalpha << right_button;
-		info["right_button"] = sstr.str();
-
-		sstr.str("");
-		sstr << mods;
-		info["mods"] = sstr.str();
-
+		info["destination"] = boost::lexical_cast<std::string>(destination);
+		info["right_button"] = boost::lexical_cast<std::string>(right_button);
```
CelticMinstrel, line 489: Okay, but doesn't this return to representing true/false as 1/0?

NQNStudios: I guess it does. After some searching, I don't see very good options for getting lexical_cast to do "true"/"false", except maybe this: https://stackoverflow.com/a/17708833

CelticMinstrel: That was the same answer I found… the problem with it is that we'd have to make a policy to never include lexical_cast.hpp directly.

We could either:

1. Do like the scenario format – save using `b ? "true" : "false"` and load as `str == "true"` (meaning any other invalid value ends up as false).
2. Create a custom function that's basically lexical_cast but deals in true/false instead of 1/0. Probably use it in the scenario format too.

NQNStudios: It'd be trivial to make 2 global inline functions out of the snippets from your option 1. I think that's what we should do--it seems kind of absurd to do anything more complicated than that.

NQNStudios: Although now I can't think of a surefire way to find all the instances of lexical_cast<string>(bool) conversions and change them. for string->bool I can grep to find every instance, but not the other way around.

CelticMinstrel: Did you manage to find them all? If it's so easy to find the casts _to_ bool then I think it should probably suffice to find the opposite cast for each of those…

NQNStudios: I just went through every boost::lexical_cast<std::string> and manually identified bool cases.

src/dialogxml/dialogs/dialog.cpp
```diff
@@ -550,7 +550,7 @@ void cDialog::handle_events() {
 			dynamic_cast<cTextField&>(getControl(currentFocus)).handleInput(key, true);
 		}else if(replaying && has_next_action("handleTab")){
 			Element& next_action = pop_next_action();
-			handleTab(boost::lexical_cast<bool>(next_action.GetText()));
+			handleTab(str_to_bool(next_action.GetText()));
```
CelticMinstrel, line 553: Am I missing it? I don't see the `bool_to_str` match for this one.

NQNStudios:
it should be here and i missed it:

https://github.com/calref/cboe/blob/b484e06fee3105e42c38436e9a6a6b97c83488a6/src/dialogxml/dialogs/dialog.cpp#L576

NQNStudios:
Now I grep 7 instances of str_to_bool and 7 instances of bool_to_str.

PR #419, opened 2024-08-28
switch the order of 2 stupid things

NQNStudios:
This fixes it for me.


NQNStudios:
I hate computers

CelticMinstrel:
Why on earth would that make a difference!? But whatever, if it works then that's all that matters.

PR #425, opened 2024-09-05
Record extra

### LINKED ISSUE #424: More to record
- [x] closing the automap
- [x] wait
- [x] act in combat
- [x] selecting a monster in the monster menu
- [x] parry

NQNStudios:
Close #424

### DIFF
--- a/src/game/boe.actions.cpp
+++ b/src/game/boe.actions.cpp
@@ -2035,6 +2035,15 @@ void debug_heal_plus_extra() {
 	put_pc_screen();
 }
 
+void close_map(bool record) {
+	if(record && recording){
+		record_action("close_map", "");
+	}
+	mini_map.setVisible(false);
+	map_visible = false;
+	mainPtr.setActive();
+}
+
 bool handle_keystroke(const sf::Event& event, cFramerateLimiter& fps_limiter){
 	bool are_done = false;
 	location pass_point; // TODO: This isn't needed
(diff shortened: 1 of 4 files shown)


CelticMinstrel:
Just making sure, but… does pressing 5 or space in combat also work? (I hadn't realized it was a different action from "wait".)

NQNStudios:
Ah jeez. So, the action for 5/space is handle_pause, but the behavior for 'w'/the wait button is different, and uses the logic I turned into `handle_wait`. So now both are recorded and replayed, but they're not the same (and they definitely probably should be.) Like, handle_pause cleans webs and handle_wait might not.

CelticMinstrel:
The one on W seems to be just "delay action" – it doesn't advance time, just lets your PCs act in a different order than usual.

NQNStudios:
Which reminds me I still have to go through all the replay actions and make
sure they only advance time if they’re supposed to.

CelticMinstrel:
I'm not clear on whether your previous comment means there is an issue, or if this is ready to merge?

NQNStudios:
There’s a known issue (with a TODO comment) that the if-else block in
boe.main.cpp calls advance_time() after every action it replays, I set it
up that way to be more DRY as opposed to calling advance_time() manually
after so many actions. So, at some point I need to go through and make
several of the action replay blocks return early to get the behavior in
alignment with how it is in normal play. But that could be another PR.

PR #432, opened 2024-09-08
forward the args from platform-specific scons scripts

NQNStudios:
In my release build github action, the workflow yaml file calls `scons-build.sh` and `scons-build.bat` with additional arguments that we don't want in those files directly. To enable this, I've modified the two scripts to forward all of their arguments to `scons` itself. This should change nothing in the upstream CI, but will make my life easier maintaining a release build branch that's easy to rebase with the upstream.


NQNStudios:
(What I'm working towards is having all of my release build logic be contained in one commit on a branch that only creates new files, it doesn't modify any upstream files. So I can easily pull from upstream and push new release builds without a merge hassle.)

PR #434, opened 2024-09-08
SConstruct changes

NQNStudios:
Here are a few more changes to `SConstruct` that I have in my release build branch, which should be harmless to the upstream CI build (but we'll see).

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -307,10 +307,6 @@ if not env.GetOption('clean'):
 		print("There's a problem with your compiler!")
 		Exit(1)
 
-	if not conf.CheckLib('zlib' if (platform == "win32" and 'mingw' not in env["TOOLS"]) else 'z'):
-		print('zlib must be installed!')
-		Exit(1)
-
 	def check_lib(lib, disp, suffixes=[], versions=[], msvc_versions=[]):
 		if "mingw" in env["TOOLS"] and lib.startswith("sfml"):
 			lib = "lib" + lib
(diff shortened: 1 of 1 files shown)


SConstruct
```diff
@@ -185,7 +185,13 @@ if platform == "darwin":
 					dest_path = path.join(target[0].path, basefile)
 					if path.exists(dest_path):
 						break
-					Execute(Copy(dest_path, src_path))
+					# Copying .frameworks needs to preserve symlinks by using cp -a.
```
CelticMinstrel, line 188: This also makes me wonder if we should use `ditto` instead of `cp`…?

SConstruct
```diff
@@ -345,6 +341,8 @@ if not env.GetOption('clean'):
 	boost_versions = ['-1_84'] # This is a bit of a hack. :(
 	suffixes = ['-mt', f'-mt-x{env["bits"]}']
 
+	zlib = 'zlib' if (platform == "win32" and 'mingw' not in env["TOOLS"]) else 'z'
+	check_lib(zlib, 'zlib', [], [])
```
CelticMinstrel, line 345: I'm confused at how this makes it be bundled…? And it's not making it be bundled on _all_ platforms, is it?

NQNStudios:
IIRC, check_lib() adds the libs it finds to bundled_libs which is where it
later finds what to bundle for Windows. I don’t think this affects Mac
because mac uses bundle_libraries_for() instead of that logic.

PR #438, opened 2024-09-09
Fosnola's fix #101

### LINKED ISSUE #101: Town boundary width and height are swapped
Towns that do not have their boundaries drawn in a square have redrawn themselves
e.g. if it is 28x30, it becomes 30x28, and moves the party out of town when it shouldn't.

---
This is an item logged from the To-Do and was originally reported by BainIhrno. If you are familiar with or can recreate this bug, please provide details here.

NQNStudios:
I've cherry-picked and amended (to remove extraneous changes) one of the commit's @fosnola said fixes issue #101, and after testing, it seems that the fix works. (However, I can in no way guarantee that it doesn't break anything else.)

There was an additional commit fosnola mentioned, https://github.com/fosnola/cboe/commit/d63b089018b8ceee8e6126bb02b575a65c2fe867 , but I looked through that one and I can't tell how it's related to town boundaries. Also it has merge conflicts against our master branch, so I thought I'd leave it alone for now.

Edit: I also don't know how it could break legacy saves, or if that other commit was to fix that potential problem.

### DIFF
--- a/src/porting.cpp
+++ b/src/porting.cpp
@@ -10,7 +10,12 @@
 
 extern bool cur_scen_is_mac, mac_is_intel;
 
+// check endian with cur_scen_is_mac and mac_is_intel
+// check cur_scen_is_mac to see if we need to convert RECT in Rect
+static void port_rect(legacy::Rect* s);
+
 void port_town(legacy::town_record_type* dummy_town_ptr){
+	port_rect(&dummy_town_ptr->in_town_rect);
 	if(cur_scen_is_mac != mac_is_intel)
 		return;
 	flip_short(&dummy_town_ptr->town_chop_time);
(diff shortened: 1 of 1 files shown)


NQNStudios:
This bug would actually be a good candidate for having a unit test.

CelticMinstrel:
Any ideas on what that unit test could look like? We could add it in this PR as well.

NQNStudios:
It could load the legacy test scenarios and check the town boundary
rectangles directly to make assertions.

Where would you want me to put test scenarios in the repository?

CelticMinstrel:
I guess they could go in rsrc/ in a new folder, like test or test-scenarios or something like that. The other option would be in test/files/.

NQNStudios:
Closed in favor of #441

Issue #192, opened 2018-05-25
MacOS 10.13.4 Not able to run


SCasarotto:
I downloaded the Mac build here: http://spiderwebforums.ipbhost.com/forum/12-blades-of-exile/ and am unable to open the game on macOS 10.13.4. It seems to be working on 10.12.6 just fine. So it must have been a change in 10.13.X that broke it. 

Anyone have any suggestions or ways I can try to diagnose and help?

wrldwzrd89:
I experienced this issue myself. The cause is a failure to link required libraries with the main app on launch - and if it works on an older version, clearly a change there is to blame. I’m honestly not sure what the proper way to debug this is.

CelticMinstrel:
I'm unable to test on _any_ modern Mac, so I can't really do anything about this. Pull requests are welcome though, as long as they don't make the Xcode 4 project file incompatible with Xcode 4 and OSX 10.7.

wrldwzrd89:
I've forked the project and got the build (mostly) fixed on macOS 10.13. The only remaining issue is that the .dylibs aren't linked up properly, so that the final packaged build can find them. I noticed there's a script for this very purpose - fix_dylibs.sh - but SCons doesn't seem to call it anywhere, and my attempts to add it to the Mac packaging SConscript file are running into errors due to a non-existent "path" variable. Below is the code I added to the file:

`
targets = [
	"Blades of Exile",
	"BoE Character Editor",
	"BoE Scenario Editor",
]
for targ in targets:
	target_dir = path.join(install_dir, targ + '.app', 'Contents/Frameworks')
	binary = path.join(install_dir, targ + '.app', 'Contents/MacOS', targ)
	env.Command(None, binary, action='pkg/mac/fix_dylibs.sh')
`

When I try to invoke scons with this change, I get:
`NameError: variable 'path' is not defined`
...from Python. Attempting to add path to the list of imports also fails, this time telling me path can't be imported because it does not exist... but it's used in the main SConstruct build file?! I'm confused.

CelticMinstrel:
The module is actually `os.path`, so to use it as in your snippet you need `import os.path as path`.

SCasarotto:
Any progress on this?

wrldwzrd89:
I've been trying to work on it, but travel and illness have put a severe damper on available time to do so. I'll pick it back up again.

wrldwzrd89:
[...]

PR #439, opened 2024-09-09
Fix rpaths

NQNStudios:
This adds a 1-file submodule which just contains my python script that detects and fixes broken rpath/dependency references in the app bundles. And extends SConstruct to make sure the submodule is cloned, and call the script automatically.

This theoretically would have fixed the problems behind #192 although the script might not cover every possible edge case.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,3 +8,6 @@
 [submodule "deps/cppcodec"]
 	path = deps/cppcodec
 	url = https://github.com/tplgy/cppcodec
+[submodule "deps/fix-rpaths"]
+	path = deps/fix-rpaths
+	url = https://gist.github.com/NQNStudios/7145bcf6621891f5176c8caa165d6b93
--- a/SConstruct
+++ b/SConstruct
@@ -469,6 +469,13 @@ if platform == "darwin":
 		target_dir = path.join(install_dir, targ + '.app', 'Contents/Frameworks')
 		binary = path.join(install_dir, targ + '.app', 'Contents/MacOS', targ)
 		env.Command(Dir(target_dir), binary, [Delete(target_dir), bundle_libraries_for])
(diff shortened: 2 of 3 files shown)


CelticMinstrel:
1. Can we call it after Xcode builds too?
2. Is there an advantage to making this a submodule rather than just adding the file?

NQNStudios:
1. I'm not sure how to set that up. I can't get the xcode build to work either.
2. That advantage is on my end -- I'm using it for other things, and I only want to maintain it in one place.

CelticMinstrel:
> I can't get the xcode build to work either.

Oh right, I think you mentioned this before. What was the actual problem with it? Unable to find SFML or something?

NQNStudios:
Yeah it was something like that. Actually, I might be able to get it to work now using my from-source dependency builds.

.github/workflows/ci.yml
```diff
@@ -36,7 +36,7 @@
 				},
 				{
 					name: build,
-					run: ./.github/workflows/scripts/mac/xcode-build.sh
+					run: ./.github/workflows/scripts/mac/xcode-build.sh || cat ./out.txt
```
CelticMinstrel, line 39: Should this be `> ./out.txt`?

NQNStudios: The Xcode build script stage isn't printing its output, so I'm writing it to out.txt, then using `cat` here to see it in the action log.

NQNStudios:
I got it working!

CelticMinstrel:
`readlink -f` does not work. I also can't find documentation on what it's supposed to do. Can you rewrite your script to use [Path.resolve](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve) instead?

NQNStudios:
Yeah

PR #443, opened 2024-09-13
Record version info

### LINKED ISSUE #440: Record the Blades of Exile version producing the action log
This should help with debugging

NQNStudios:
Close #440 

I found a bug where scons builds defined GIT_TAG as empty because of a trailing blank line in the subprocess output.

Fixed that, and am now recording a root element like this:

```
<actions SHA="f411a46" Tag="v0.0.4" Status="modified" Repo="git@github.com:NQNStudios/cboe.git">
```

The CI will show whether my changes to `gitrev.sh` work for the Xcode build.

### DIFF
--- a/src/tools/replay.cpp
+++ b/src/tools/replay.cpp
@@ -15,6 +15,7 @@
 #include <codecvt>
 #include <string>
 #include <boost/lexical_cast.hpp>
+#include "tools/gitrev.hpp"
 #include "gfx/render_shapes.hpp"
 
 using base64 = cppcodec::base64_rfc4648;
@@ -49,6 +50,9 @@ bool init_action_log(std::string command, std::string file) {
 		log_file = file;
 		try {
 			Element root_element("actions");
+			root_element.SetAttribute("SHA", GIT_REVISION);
+			root_element.SetAttribute("Tag", GIT_TAG);
+			root_element.SetAttribute("Status", GIT_STATUS);
 			log_document.InsertEndChild(root_element);
 			log_document.SaveFile(log_file);
 			recording = true;
--- a/src/tools/replay.cpp
+++ b/src/tools/replay.cpp
@@ -15,8 +15,10 @@
 #include <codecvt>
 #include <string>
 #include <boost/lexical_cast.hpp>
-#include "tools/gitrev.hpp"
 #include "gfx/render_shapes.hpp"
+#ifndef MSBUILD_GITREV
+#include "tools/gitrev.hpp"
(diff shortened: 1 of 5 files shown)


NQNStudios:
Seems like Visual Studio builds don't generate tools/gitrev.hpp

NQNStudios:
wait, there's a ClInclude for it...

CelticMinstrel:
What's v0.0.4?

NQNStudios:
v0.0.4 is a tag in my fork.

NQNStudios:
Wow, I added this guard which is the same as other places gitrev.hpp is included:

```
#ifndef MSBUILD_GITREV
#include "tools/gitrev.hpp"
#endif
```

and it didn't work!

CelticMinstrel:
Maybe the project file is configured to only define MSBUILD_GITREV in one file?

src/tools/replay.cpp
```diff
@@ -50,10 +52,14 @@ bool init_action_log(std::string command, std::string file) {
 		log_file = file;
 		try {
 			Element root_element("actions");
+			#if defined(GIT_REVISION) && defined(GIT_TAG)
 			root_element.SetAttribute("SHA", GIT_REVISION);
 			root_element.SetAttribute("Tag", GIT_TAG);
+			#endif
+			#if defined(GIT_STATUS) && defined(GIT_REPO)
```
CelticMinstrel, line 59: Is there a reason for doing these in pairs?

NQNStudios:
I think it's because I thought the Visual Studio build had a special way of defining the values that I hadn't implemented the new ones for. But now I'm pretty sure Visual Studio builds just don't have any of the values defined. I do need to extend the no-git-repo case in the SConstruct to define blank strings for the new values though, then this can be a single ifdef

NQNStudios:
or maybe each one should have its own guard. i don't know.

NQNStudios:
@CelticMinstrel Is there anything else you want to see from this one?

CelticMinstrel:
Hmm… does the Visual Studio build have this information at all? Does it show the commit hash on the startup screen?

NQNStudios:
It does not.

CelticMinstrel:
That is _quite_ unfortunate… but I'm not sure fixing that is in scope for a PR whose purpose is just to add that info to the replay…

It's possible to code the generation of this information directly into the Visual Studio project file. I'm not sure if this approach would survive saving the project from Visual Studio, though.

PR #445, opened 2024-09-16
fix-rpaths use Path.resolve() instead of readlink -f

NQNStudios:
Hopefully this works


CelticMinstrel:
This builds and runs on my machine. I can't easily verify that it's portable (ie, will run on a non-dev machine), but I'll assume it is for now.

PR #448, opened 2024-09-19
Fix scons tests

NQNStudios:
I've been trying to pave the way for #441 by making the scons ci builds fail when the tests fail.

This proved troublesome because the Linux and Win-Scons builds have been failing the tests silently for quite a while.

For some reason, Linux needs an X server to run the tests -- this was a pretty trivial fix because a utility is included on Ubuntu github actions runners called `xvfb` which I had never heard of but seems really awesome. It runs a headless X server for CI workflows. Using this fixed that problem.

(Note for the future: there's a github action that lets you use `xvfb` on Windows and Mac. Maybe we could someday even be running our full replay test suite on CI for all three platforms this way. I don't know how sophisticated it is.)

Now, for win-scons the tests are failing with an error code which I think indicates missing DLLs. I've tried so many different things to fix it, but I can't repro this failure on my own machine. On my PC, the tests run and succeed through Scons. I'm at my wits' end and I'm starting to think it's a poor use of time to keep making this a blocking issue. So I would suggest that we should make the win-scons workflow succeed if it compiles correctly, but issue a warning that the tests didn't run. @CelticMinstrel would this be okay with you, alongside opening a github issue for the real problem?

I'm also trying to trim down the other Github actions warnings, so that the win-scons warning (which I haven't implemented yet) will stand out.

### DIFF
--- a/.github/workflows/scripts/linux/scons-build.sh
+++ b/.github/workflows/scripts/linux/scons-build.sh
@@ -1,0 +1,3 @@
+#!/bin/sh -ve
+
+CCFLAGS=-fdiagnostics-color=always scons
--- a/.github/workflows/scripts/linux/scons-build.sh
+++ b/.github/workflows/scripts/linux/scons-build.sh
@@ -1,3 +1,4 @@
 #!/bin/sh -ve
 
-CCFLAGS=-fdiagnostics-color=always scons
+export CCFLAGS=-fdiagnostics-color=always
+scons
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,7 +155,6 @@
 					name: build and unit test,
 					run: '.\.github\workflows\scripts\win\scons-build.bat'
 				},
-
 			]
 		},
 		# win-mingw: {
@@ -198,10 +197,10 @@
 					run: 'sudo ./.github/workflows/scripts/linux/install-tgui.sh'
 				},
 				{
-					name: build and unit test,
-					run: CCFLAGS=-fdiagnostics-color=always scons
+					name: build and unit test headless,
+					run: 'xvfb-run ./.github/workflows/scripts/linux/scons-build.sh'
 				}
-			],
+			]
 		}
 	}
 }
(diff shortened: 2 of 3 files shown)


.github/workflows/scripts/linux/scons-build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh -ve
+
+CCFLAGS=-fdiagnostics-color=always scons
```
CelticMinstrel, line 3: Since this is now in a separate file, I think it should go on two lines:

```suggestion
export CCFLAGS=-fdiagnostics-color=always
scons
```

I'm _pretty_ sure that would have the same effect.

CelticMinstrel:
> For some reason, Linux needs an X server to run the tests -- this was a pretty trivial fix because a utility is included on Ubuntu github actions runners called `xvfb` which I had never heard of but seems really awesome. It runs a headless X server for CI workflows. Using this fixed that problem.

Feels like it's potentially a bug that the tests need an X server, but I guess it doesn't really matter that much – it's more important that the tests run at all.

> Now, for win-scons the tests are failing with an error code which I think indicates missing DLLs. I've tried so many different things to fix it, but I can't repro this failure on my own machine. On my PC, the tests run and succeed through Scons. I'm at my wits' end and I'm starting to think it's a poor use of time to keep making this a blocking issue. So I would suggest that we should make the win-scons workflow succeed if it compiles correctly, but issue a warning that the tests didn't run. @CelticMinstrel would this be okay with you, alongside opening a github issue for the real problem?

Have you checked to make sure that the current working directory that scons uses to run the tests is the directory that contains the dlls? If you've already covered that avenue, then sure, we can open an issue and defer it for now.

NQNStudios:
This looks to be working. The last two commits could be squashed together but I'm leaving them for now to show why the double-quote syntax doesn't match the rest of ci.yml.

NQNStudios:
Ready for this to merge if you are, @CelticMinstrel

PR #452, opened 2024-09-22
Fix #451

### LINKED ISSUE #451: [linux] scenario editor new scenario file picker doesn't allow creating new file
> the Zenity picker doesn't let you choose a file that doesn't exist yet, so creating a new scenario in the editor is impossible (maybe you could do it by picking a file that DOES exist an overwriting it, but... that would be bad...

The New Game file picker in the core game does, so this fix should be trivial.

NQNStudios:


CelticMinstrel:
This isn't really related, but I just realized that the `filename += buffer` line might be unsafe. I'm not sure if it correctly accounts for the fact that the buffer is known to be 128 characters.

NQNStudios:
That's here, for conversational reference:

https://github.com/calref/cboe/blob/f4c34a410d04d53106ee0f230cffc5c859e7810b/src/tools/winutil.linux.cpp#L146

Also, is it reading chunks of 128 characters at a time from a command? Why not just read 1 at a time until it reaches the end rather than request more than is known to be there?

CelticMinstrel:
I don't really know. I assume that requesting 128 characters when there are only 30 characters is fine though.

NQNStudios:
So, what would make it safe? Setting the index after the end of the buffer to the zero/null terminator? Constructing a std::string before calling += ?

CelticMinstrel:
I was thinking using `std::append` instead of `+=`, as there's a form of that which takes a count. But then I looked at the `fgets` documentation and realized that it actually reads only 127 characters and adds a null terminator, so `+=` seems to be safe after all.

Issue #455, opened 2024-09-24
"Ask About" text input crash


NQNStudios:
I chose "Ask About" in dialog, typed nonsense into the text field, and pressed enter. The game crashed.

Edit: It happened when talking to Avizo, but it *didn't* happen when talking to Hamer! (just like with #454)

PR #457, opened 2024-09-25
Record advance time

### LINKED ISSUE #437: When replaying, match the original advance_time() behavior
This is that known bug in the replay system that I need to get around to fixing.

NQNStudios:
Fix #437

* Record and replay actions from #422
* Implement `--verbose` flag for replay recording, which records internal testing actions (for now just `advance_time`)
* This PR I think covers making the vast majority of action cases perform advance_time() identically to original runtime behavior. I would say it's highly likely that I missed edge cases, but this version of the code is much more correct.
* Also of note: obsessively matching the behavior of `advance_time()` while correct and good, is not entirely necessary for replay integrity as long as action replays always handle the `did_something` flag correctly.

To test this and find every discrepancy, I used these two replays which together contain one of every action (I'm pretty sure):

[Archive.zip](https://github.com/user-attachments/files/17137020/Archive.zip)

There are some major caveats to those 2 replays, which is why I am not staging them in the repository:

* They were both recorded with different builds, each in a modified state
* In the first one, I deleted some actions at the end which led to the crash #455
* As I was testing and fixing, I changed the way some actions record themselves. Rather than re-record the whole logs, I would edit the log to match the changing behavior (but I could have made a mistake).
* They take very very long.

Some things about these replays are cool:
* They recreate pretty long play sessions, appearing to match the state of the original playthrough perfectly, including in combat (which is fun to watch)
* You can see the replays reproduce some UI bugs which would be impossible to check with unit testing
* If I ever demo OpenBoE at a public event, I'll probably make replays to have looping when no one's actively playing it :)

This is a wibbly wobbly PR which I'm afraid rebasing for cleaner history, would get too hairy. So I haven't tried squashing anything yet. I also don't think it would be proper to squash the whole PR into one commit, because it shows the evolution of the `--verbose` flag and my reasoning for it. All the contiguous "fix some cases" commits could probably be squashed together easily, though.

My next recording system PRs after this one will do the following to address the dangling problems:

* Add two folders, `test/replays/short` and `test/replays/long`
* Add a much shorter log that records one of everything FROM A CLEAN REPO STATE, WITHOUT MANUAL MODIFICATIONS
* Add a bash script `test/replays.sh` which runs the replays, with the options to fail fast and/or skip the long ones.
* On success of a comprehensive run of all replays, this script will update a file `test/replays/success_{platform}.txt` which records the latest gitrev info unless the repository has been modified. This file can be staged, and potentially the CI for release builds could check for its existence and assert that a pull request comes with a recent success SHA for all 3 platforms (though this requires a dev to have access to all of Mac/Windows/Linux, which I do for now, but obviously not everyone will and my computers might break some day.)
* Get the replay test to pass on all 3 platforms so that `success` files can exist for all 3.


NQNStudios:
That win-scons one failed for github action reasons. Needs to re-run

CelticMinstrel:
> though this requires a dev to have access to all of Mac/Windows/Linux, which I do for now

I technically have access to all 3 too, though my Windows/Linux computer has been turned off for quite awhile and probably needs servicing.

src/tools/replay.cpp
```diff
@@ -156,7 +156,10 @@ Element& pop_next_action(std::string expected_action_type) {
 		replay_fps_limit->frame_finished();
 	}
 	
-	last_action_type = to_return->Value();
+	// control_click actions are not meaningful for debugging
```
CelticMinstrel, line 187: What is this?

src/game/boe.specials.cpp
```diff
@@ -1282,6 +1282,7 @@ bool use_space(location where) {
 // Note ... if this is a container, the code must first process any specials. If
 // specials return false, can't get items inside. If true, can get items inside.
 // Can't get items out in combat.
+// TODO this always returns false. Why?
```
CelticMinstrel, line 1285: I suspect it's supposed to return `s1`.

src/tools/replay.cpp
```diff
@@ -159,6 +187,12 @@ Element& pop_next_action(std::string expected_action_type) {
 	// control_click actions are not meaningful for debugging
 	if (to_return->Value() != "control_click"){
 		last_action_type = to_return->Value();
+
+		if(replay_verbose){
+			// Verbose replays are for internal testing, so this console output won't bother
+			// anyone and is very helpful:
+			std::cout << last_action_type << std::endl;
```
CelticMinstrel, line 194: I think this should have a common prefix, like "Replayed action:" or something.

src/game/boe.main.cpp
```diff
@@ -602,13 +602,14 @@ static void replay_next_action() {
...
 		handle_toggle_active(need_reprint);
-		return;
 	}else if(t == "handle_parry"){
 		handle_parry(did_something, need_redraw, need_reprint);
 	}else if(t == "handle_monster_info_menu"){
 		handle_monster_info_menu(boost::lexical_cast<int>(next_action.GetText()));
+		return;
 	}else if(t == "cancel_item_target"){
 		cancel_item_target();
+		return;
```
CelticMinstrel, line 619: This seems wrong (though in this case it's that I probably should've _made_ it advance time and didn't). Assuming time advances when casting Identify outdoors, it should also advance when calling Identify in town. (The item target mode is only used in town.)

CelticMinstrel:
This one looks wrong: 762feecd2ba4bf9736b71180db4ad45385d9bf04

One other that looked wrong is `handle_alchemy`, but I'm not sure of that – I think it _should_ advance time if you make a potion, but it certainly shouldn't advance time if you open the alchemy menu and then cancel. Were both those cases tested?

There are a couple of TODO notes that need removing now. One is the one you can see in 47b4456082d0eb530366e494fb18f270679e37ff and the other is the original TODO note about validating advance_time.

NQNStudios:
> This one looks wrong: https://github.com/calref/cboe/commit/762feecd2ba4bf9736b71180db4ad45385d9bf04

I believe it's correct -- screen shifting needs to call advance_time() with need_redraw true but did_something false

NQNStudios:
> One other that looked wrong is handle_alchemy, but I'm not sure of that – I think it should advance time if you make a potion, but it certainly shouldn't advance time if you open the alchemy menu and then cancel. Were both those cases tested?

Oh, I was definitely rushing though when I recorded handle_alchemy. I have to redo that one and make sure to get both cases

CelticMinstrel:
> > This one looks wrong: [762feec](https://github.com/calref/cboe/commit/762feecd2ba4bf9736b71180db4ad45385d9bf04)
> 
> I believe it's correct -- screen shifting needs to call advance_time() with need_redraw true but did_something false

Ah, I guess that half makes sense. It suggests `advance_time` isn't _quite_ the right name for it, but close enough I guess.

> Oh, I was definitely rushing though when I recorded handle_alchemy. I have to redo that one and make sure to get both cases

What about other dialog things that can be cancelled? Like casting a spell, or deleting a PC…

src/tools/replay.cpp
NQNStudios: Tracking the last action to be replayed, which is helpful in the error messages for incorrect advance_time behavior: you want to know which command replayed last (that's the incorrect one)

CelticMinstrel: That wasn't the question. I'm asking what `control_click` is.

NQNStudios: `control_click` is an action recorded by a dialog, which means a control with the given id got triggered. (Keystrokes  that trigger controls' click events also record this event.)

They're not meaningful as a last_action_type because they get picked up and processed by the dialog's internal event loop when it runs.

NQNStudios:
No, I didn't comprehensively test canceling cancelable actions. That actually is probably gonna reveal new bugs and discrepancies

NQNStudios:
> Ah, I guess that half makes sense. It suggests advance_time isn't quite the right name for it, but close enough I guess.

Right. `advance_time` only actually advances time if an action set `did_something` to true. The other two flags just redraw the screen and print out buffer messages. `did_something` is very rarely set *separately* from the business logic of actions' functions in `boe.actions.cpp`, which means we possibly had leeway to keep calling advance_time after *every* action in the first place.

Unless `advance_time()` changed behaviors.

NQNStudios:
> Because --verbose just records what happens. Actually I would expect those cases to just work.

I can express this better.

Actions are calling synchronous dialogs to either complete the action or cancel it. *After* the synchronous dialog runs, they're calling advance_time.

At replay-time, the dialogs also run synchronously and complete the action logic (accounting for cancel) before *checking* advance_time. So it should match without extra work.

src/tools/replay.cpp
CelticMinstrel: Ah, it's a poor choice of name, I guess. It sounds like holding the control key and pressing a mouse button. Would probably be less confusing as `click_control`.

NQNStudios:
https://github.com/calref/cboe/pull/457/files#diff-0fb416d013a449a9dcaaf68891fc4bf89204fbb9a69ea747e6c27141026ea24aL424

if this ever triggers a spell being cast outside of combat, it should probably advance time. Inside of combat, I assume the Action Points system would handle advancing the time.

test/replays/short/Issue455.xml
```diff
@@ -0,0 +1,65 @@
...
+UIScale = 1.5
+UIScaleMap = 1.5
+]]>
+</load_prefs>
+    <srand>1727292388</srand>
+    <startup_button_click>
+        <btn>3</btn>
+        <mods>1</mods>
+    </startup_button_click>
+    <build_scen_headers>AnimatedTile.boes&#x0A;Boundary Test.boes&#x0A;Boundary Test 2.boes&#x0A;Display Picture.boes&#x0A;myscenario.boes&#x0A;thrakos.exs&#x0A;recttown.exs&#x0A;</build_scen_headers>
```
CelticMinstrel, line 25: Could you please make this one use a CDATA section, like I did with the prefs?

CelticMinstrel:
How is the base64 content formatted when a load game action is recorded? If possible I'd like it to be broken into fixed-length lines (except for the final line). It could maybe use a CDATA section… I think that might be required to preserve the line breaks with the way you've implemented the replay system, though this isn't really a case where whitespace needs to be preserved at all… but a base64 decoder is supposed to ignore whitespace anyway so it shouldn't matter.

CelticMinstrel (commented):
Are you planning to finish this anytime soon?

NQNStudios:
I’m coming back from an RSI-induced hiatus and my efforts are a little
diffuse across different issues right now. So yes, but I’m distracted.

NQNStudios:
> How is the base64 content formatted when a load game action is recorded? If possible I'd like it to be broken into fixed-length lines (except for the final line). It could maybe use a CDATA section… I think that might be required to preserve the line breaks with the way you've implemented the replay system, though this isn't really a case where whitespace needs to be preserved at all… but a base64 decoder is supposed to ignore whitespace anyway so it shouldn't matter.

The Linux man page for base64 says 76 is the default chars-per-line of base64 output:
https://linux.die.net/man/1/base64

I can't find any definitive standard about how many chars per line there should be, so I'll use 76.

PR #458, opened 2024-09-26
Auto-generated credits / Update the credits from IndieGOGO

### LINKED ISSUE #377: Credits overhaul
The in-game credits live here: [startanim.png](https://github.com/calref/cboe/blob/master/rsrc/graphics/startanim.png)

A ton of past contributors are missing from these credits, plus 40 new financial supporters.

I'd like a robust way to generate this image from a text file if there's not one already. At a very low-tech level, this could just be an ImageMagick script we run manually when we need to update the animation.

I also think it would be good to have this animation, when clicked on, expand into its own dialog where a taller viewing window of the scrolling is visible, possibly also scaled up.

NQNStudios:
Fix #377 / Fix nqnstudios#1

First of all, I've made a file, ~~DONORS.md~~ `pkg/credits/Funding.txt`, which contains the **confirmed** names of every IndieGoGo backer who responded so far. There's also `pkg/credits/Programming.txt` and `pkg/credits/Graphics.txt`.

~~I'm going to email every backer and make sure they get a chance to verify their name, change it, and if they didn't select the reward, maybe it was an accident, so I'd add those names later.~~ I've emailed the backers several times for name confirmations. Only 5 backers have not yet confirmed. I've pulled their names and will add them later when I get ahold of those folks.

The script, `pkg/generate-credits.py` (which depends on ImageMagick) must be run manually whenever we change the names in the credits.

This script generates 2 things:

* A startanim.png that stitches together the original credits and new OpenBoE credits
* rsrc/dialogs/about-boe.xml containing all of the new OpenBoE credits. `pkg/credits/about-boe-template.xml` defines how the generated dialog will look.

So if the credit names change, or about-boe-template.xml changes, we run the script again. And we stage the files that it changes.

I think it's better for this to be manual, not integrated with the build, because ImageMagick shouldn't be required for building the code, and integrating scripts into 3 different builds (scons, visual studio, xcode) is not worth the time and nightmare it take.

I've created #469 to track more names we need to hunt down to credit all the programmers and artists that have been missed.

If you don't hate the approach that I've taken/the look of the new startanim.png, then this should be ready to merge and close the issue. It handles all the meat of the overhaul.


CelticMinstrel:
Definitely need to verify exactly what name they prefer to be referred to by. Maybe there are some who'd like to use a pseudonym instead of (or in addition to) their real name.

With that many names, it seems iffy to put them in the scrolling credits (though I wouldn't say it's _impossible_). Regardless of whether they go in the scrolling credits, they should also go in the about dialog… which probably means the about dialog will need to be cleaned up to avoid the "two overlapping text blocks that properly line up based on mutually exclusive whitespace" mechanic.)

NQNStudios:
> I don't really like the idea of parsing this file to pull out the bullet points, but I guess there are worse ways to do it? There is also the possibility of parsing it at build time though, and dumping the output into the strings folder.

The approach I'm working with now, is to parse the file and generate the image and DialogXML from a bash script that we will have to run manually when we change credit metadata. Then we stage the generated files.

I think manually is better than build-time in this case, because every time we add a build-time step, we have to implement it for Scons, Xcode, *and* Visual Studio, which is a nightmare every time.

NQNStudios:
@CelticMinstrel The start animation with added credits doesn't look *great*, but I think I'm all done with the automation required for us to extend the about menu and start animation with lots of missing open-source credit information.

I started collecting this in `pkg/credits/*.txt` for each type of contribution. Each text file can have a Not Confirmed: heading where I've put backers who haven't confirmed yet, but also where I've put repo contributors who might want their real name instead of GitHub username. And I know that more people might have contributed art that only you would know how to try getting in touch with them.

NQNStudios:
I cleaned up the git history and edited the PR description to cover most everything.

NQNStudios:
Actually--it's not gonna be very hard to make the credits font match, even without knowing what the original font was. Enough letters are present to extract a bitmap font from it, and only be guessing at how a couple letters might have looked.

NQNStudios:
@CelticMinstrel I've just gone and made a proper bitmap font out of the startanim letters that we have. The only ones that I had to create myself were `QXYZ`, `0`, `()`, `jq`, and I think they look pretty good and fit in. If we get more contributors with numerals in their names, accent marks, etc. we will have to expand the font. I made it with FontForge on Mac.

This makes the generated startanim look way better, more cohesive, and I think turns this PR into all upside.

NQNStudios:
image

Now that I think of it, I think I want to change this so it says "Created 1997" and remove the "All Rights Reserved". Maybe instead of "All Rights Reserved", "Free and Open Source", or "Community-owned" or something although this scripty font is trickier to mess with than the smaller one and has less baseline letters to work with.

NQNStudios:
image

CelticMinstrel (commented):
Various thoughts in no particular order…

* I guess the bitmap font is fine. It's too bad we can't find the original font, but whatever.
* Stitching the startanim as a build step is an interesting idea. I'd entertained the possibility of composing on the fly at runtime, but I guess this works.
* I don't think there's a need to save a copy of the "original" startanim. It's not even the original one – it's been modified several times.
* In startanim-pt-2 it looks like there's enough room to show the names in two columns.
* It's not like I have an issue with using ImageMagick, but…
  * if you're doing it in Python it kinda makes more sense to use pillow instead? Though I'm not sure if pillow can actually do the things you've done there.
  * Even if pillow won't work or is too much effort, can it at least be done without requiring git bash on Windows?
  * Using `cat` to receive keyboard input is kind of terrible… though I guess it's being piped in programmatically so at least it's not confusing for the user.
* There sure is some excessive indentation in the generated about-boe.xml? This isn't really a big deal, but it might be nice if it can be avoided.
* It seems you've added a new `align` attribute, so please update the [dialogxml documentation](https://github.com/calref/cboe/blob/master/src/doxy/mainpage.md).
* Can we have a better way of detecting "this dialog has hyperlinks" than "it's the about dialog"? Though I suppose that could be done in a separate PR. There already is at least one other dialog that has hyperlinks though, does it not go through that particular code path?
*

NQNStudios:
> I guess the bitmap font is fine. It's too bad we can't find the original font, but whatever.

We could ask JV if he remembers, and maybe get access to the full font later. I bet he won't be quick in responding, likely forgot, and/or the answer could be "it was a custom bitmap font to begin with and I lost the file"

NQNStudios:
> In startanim-pt-2 it looks like there's enough room to show the names in two columns.

The Almighty Doer of Stuff's name is the only one that goes past 50% of its width. Splitting it into 2 lines with the second one indented, would make 2 columns possible which I agree would be better.

NQNStudios:
> Can we have a better way of detecting "this dialog has hyperlinks" than "it's the about dialog"? Though I suppose that could be done in a separate PR. There already is at least one other dialog that has hyperlinks though, does it not go through that particular code path?

The other dialog with hyperlinks was the only one, and handled specially. I've gone ahead and made it so all hyperlinks are wired up automatically.

src/dialogxml/widgets/control.hpp
```diff
@@ -32,6 +32,9 @@ namespace ticpp {
 	class Node;
 }
 
+// Hyperlink forward declaration
+void launchURL(std::string url);
```
CelticMinstrel, line 36: This could go in the cpp, right? Should probably also have `extern`.

CelticMinstrel (commented):
Looking at the generated result, it feels that the second line of ADoS's could be indented a bit more.

CelticMinstrel:
> When I changed the header to "Created 1997, free open source", it eliminated some examples of letters in that specialized font. We might wanna hold onto those.

Ah, fair point.

> We could ask JV if he remembers, and maybe get access to the full font later. I bet he won't be quick in responding, likely forgot, and/or the answer could be "it was a custom bitmap font to begin with and I lost the file"

Even if he remembers, I can't guarantee that it hasn't been changed by us at some point.

Issue #462, opened 2024-11-01
"Cannot open include file: 'clara.hpp': No such file or directory" when compiling with VS 2017


megablinken:
I've gotten the "Cannot open include file: 'clara.hpp': No such file or directory" when compiling with Visual Studio on two different Windows PCs. The first was a Win11 machine with Visual Studio Community 2022, and I thought maybe the newer version of VS was causing problems, so I tried again with a different Win10 machine with VS 2017, but I got the same error.

Here are my setup steps:

<<< INSTALL VC2017 >>>
visual studio community 2017, installer downloaded from here:
https://aka.ms/vs/15/release/vs_community.exe

Workloads:
select Desktop development with C++  - left default settings selected, except
Added:
Windows 10 SDK (10.0.16299.0)

<<< INSTALL VCPKG >>>
(open cmd window)
cd \
git clone https://github.com/microsoft/vcpkg.git
cd vcpkg
bootstrap-vcpkg.bat -disableMetrics
vcpkg.exe integrate install

Tried with and without adding vcpkg dir to the system PATH (and VPKG_ROOT). No difference.

<<< INSTALL VCPKG LIBS >>>
Tried installing the libs in manifest mode, and without manifest mode. No difference.

Without manifest mode, I installed them using this command:
vcpkg install --triplet=x86-windows zlib sfml opengl boost-any boost-dynamic-bitset boost-ptr-container boost-core boost-filesystem boost-system boost-date-time boost-chrono boost-math boost-spirit

It seems to also need cppcodec, so I installed it also:
vcpkg install --triplet=x86-windows cppcodec

<<< INSTALL BOE SRC >>>
(open cmd window)
cd \
md BoE
cd BoE
git clone https://github.com/calref/cboe.git

<< COMPILE >>
Launch Visual Studio
Open C:\BoE\cboe\proj\vs2017\Blades of Exile.sln
Select Debug and x86
Unload everything but Common and BoE
Build

Always get this error:
Severity	Code	Description	Project	File	Line	Suppression State
Error	C1083	Cannot open include file: 'clara.hpp': No such file or directory	Blades of Exile	c:\boe\cboe\src\tools\cli.hpp	4

It looks like clara is expected to be part of the Catch namespace, but none of the catch* packages available for install via vcpkg fix this. 
I tried installing "catch2", "catch", and "catch-classic" (in this order):
vcpkg install --triplet=x86-windows catch2
vcpkg install --triplet=x86-windows catch
[...]

PR #464, opened 2024-11-05
Update README.md to address #462

NQNStudios:
Could be made more concise, and might not be git beginner-friendly enough, but I made some edits that should make our recurring problem stated in #462 less common by being extremely explicit and using bold/capitalized text and a heading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,8 +88,9 @@ For Linux builds, the following additional dependencies are required:
 If you are using the Visual Studio toolset, we recommend installing
 [vcpkg](https://github.com/Microsoft/vcpkg) to manage these dependencies.
 Use the following commands to install the required dependencies.
-**Make sure your working directory is the repository root, not the Visual
-Studio project folder.**
+**You can't use the vcpkg installation that comes with Visual Studio for
+these commands. Also make sure your working directory is the repository
+root, not the Visual Studio project folder.**
 
 For 32-bit builds:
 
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
> You can't use the vcpkg installation that comes with Visual Studio for
these commands.

…huh? Why not?

NQNStudios:
Vcpkg has two modes: manifest mode (which we use in the CI to lock the
dependency versions) and classic mode (which we are telling people to use
here). In Visual Studio 2022 Community edition I found out yesterday that
they’ve disabled classic mode in the built-in vcpkg.

So an alternative here would be to change this instruction to tell them to
use manifest mode. For some reason, we didn’t make that change in the docs
when we made it in the CI. I don’t remember if there was a good reason.

CelticMinstrel (commented):
> So an alternative here would be to change this instruction to tell them to
> use manifest mode.

What does that actually entail? How would the install commands differ?

NQNStudios:
As it happens, I think the install command would get simpler, with a caveat that you must call it from inside proj/vs2017 or proj\vs2013.

`C:\Users\natqu\repos\cboe\proj\vs2017>vcpkg install --triplet=x64-windows`

And this would help us avoid new contributors getting a new version of boost that scons can't find.

NQNStudios:
I'm gonna go ahead and make that change.

PR #466, opened 2024-11-10
fix #465

### LINKED ISSUE #465: Intro/"Welcome to Blades of Exile" dialog has broken formatting
Screenshot 2024-11-07 at 12 40 10 PM

NQNStudios:
The special `|` was getting stripped from the message's text *before* passing it to the line break calculator function, which requires the `|`s to measure correctly.

This fixes the problem!


CelticMinstrel (commented):
It's good that this fixes it, but I'm a bit confused by something else I see here… why are the pipes being replaced by spaces when they're supposed to represent a line break?

NQNStudios:
The information that the break needs to be inserted at a certain character
index is represented in the break_info struct produced by
calculate_line_wrapping()

Issue #454, opened 2024-09-24
Text invisible in certain text fields (mac)


NQNStudios:
Here's a new one:

<img width="434" alt="Screenshot 2024-09-24 at 10 14 06 AM" src="https://github.com/user-attachments/assets/fdc212ff-605b-44bb-ba73-4a8c56afa0c2">

The text is logically there, just not drawn.

NQNStudios:
This also happens in the 'Ask About' text field. Maybe it happens with every text field.

NQNStudios:
I was wrong. Not fixed.

NQNStudios:
I think this was somehow related to #479 because they were both happening at the same time in the replay of that bug.

I've just gone around testing "Ask About" with a ton of different NPCs and not gotten the bug. And a ton of text fields in the scenario editor. I've been wrong before but since I can't reproduce this anymore, I'm gonna say it's dead for real, knock on wood.

NQNStudios:
It happened to me yesterday, with both the text buffer and text fields getting messed up invisible characters at the same time. Too bad I wasn't recording a replay.

Issue #427, opened 2024-09-07
The "don't understand" response doesn't seem to work.


CelticMinstrel:
For example, ask "Commander Terrance" in Fort Talrus about "bugs". The result is a blank response, instead of "He looks confused".

NQNStudios:
I was thinking this might be happening for similar reasons to #420 https://github.com/calref/cboe/pull/421/commits/051130a4e1b293dcbf02842600fb3c20890a680e

CelticMinstrel:
Fixed by #467

PR #467, opened 2024-11-10
Fix #455

### LINKED ISSUE #455: "Ask About" text input crash

NQNStudios:
#455 was very possibly a Windows manifestation of undefined behavior that also causes #454. I've got to check on Mac to see if this PR fixes both.

### DIFF
--- a/src/gfx/render_text.cpp
+++ b/src/gfx/render_text.cpp
@@ -101,10 +101,8 @@ break_info_t calculate_line_wrapping(rectangle dest_rect, std::string str, TextS
 	str_to_draw.setString(str);
 	short total_width = str_to_draw.getLocalBounds().width;
 
-	if(total_width < dest_rect.width()){
-		// The text fits on one line, so break_info won't end up being used by win_draw_string anyway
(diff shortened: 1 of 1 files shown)


NQNStudios:
Ok. This PR also appears to fix both #454 and #427.

These bugs were all caused by calculate_line_wrapping() short-circuiting when the text fits on one line. That function actually can't short-circuit because it serves a dual-purpose of calculating word bound rectangles.

I can't currently explain why some NPCs were affected by #454 and #455 but other NPCs weren't. It should have been the case that every one-line text message calling draw_string_hilite should have been affected--and Hamer's "don't understand" response is one line just like Avizo's. I noticed in the debugger that Avizo's "don't understand response" ends up being processed with a `|` at the end of it. Maybe Hamer's doesn't--I didn't debug the Hamer case because it seems fixed by the simple change.

PR #472, opened 2024-11-18
Pick the largest suitable default scale. Fix #468

### LINKED ISSUE #468: Default graphics/UI scale >1?
I think given modern monitor sizes, players are more likely than not to have extra space for the game window. Launching it at 1x scale can actually put strain on the eyes--and people might not think to go to the preferences to fix that. Maybe we could make the default setting 1.5x or even 2x?

NQNStudios:


src/pcedit/pc.graphics.cpp
```diff
@@ -29,7 +29,8 @@ extern fs::path file_in_mem;
 extern short store_flags[3];
 extern short current_active_pc;
 
-rectangle whole_win_rect = {0,0,440,590};
+// TODO why are width and height switched here?
```
CelticMinstrel, line 32: What do you mean, they're switched? The rectangle constructor is top, left, bottom, right.

src/tools/winutil.hpp
```diff
@@ -62,6 +66,44 @@ inline int os_specific_y_offset() {
 #endif
 }
 
+// The default scale should be the largest that the user's screen can fit all three
+// BoE application windows (because they should probably default to match each other).
+inline float fallback_scale() {
+	static float _fallback_scale = 0;
```
CelticMinstrel, line 72: 1. I don't like variable names beginning with an underscore. Wouldn't just `scale` do just fine?
2. This is a pretty big function, does it really need to be defined inline in the header?

src/pcedit/pc.graphics.cpp
NQNStudios: I'm really used to the order being {x, y, width, height} in other frameworks. My bad.

src/tools/winutil.hpp
NQNStudios: 1. Yes
2. Probably not. I just need to find a different header/source file that is included in the Common sources where it would make sense to be forward-declared/defined. Because if I declare it in winutil.hpp, the definition would have to be repeated in the 3 different winutil.*.cpp/mm files.

Issue #476, opened 2024-11-21
Modifier release events dropped when starting scenario


NQNStudios:
In the current master branch, on Mac, I'm experiencing a game-breaking bug where entering look mode seems to be permanent. I can click to look at lots of tiles, search a dresser and get items, but I never return to movement mode.

Related: I wish #322 was also fixed.

NQNStudios:
https://github.com/calref/cboe/blob/b2c0af3d48195eed08b1b03a3a1d862943929eab/src/game/boe.actions.cpp#L546-L547

Look mode is getting set false here. Which indicates to me that this a Mac-specific keyboard input bug. I'm not pressing any modifier keys.

NQNStudios:
Oh, you know what? I am launching VoDT with an option-click. So that release event must not be getting fired.

NQNStudios:
Something about the transition between the main menu and launching the scenario must be dropping the release event, because when I hold option in-game, the state doesn't get stuck.

NQNStudios:
Maybe it's the startbtn event loop!

NQNStudios:
This bug still affects Windows. Weird.

PR #477, opened 2024-11-21
wrap pollEvent() to guarantee all loops handle modifiers

NQNStudios:
This was meant to address #476 but doesn't actually solve the problem. Still might be a good change, though--

We're calling pollEvent on a lot of windows all over the place, and I see a lot of places (maybe only for a few frames, but still) where events never seem to reach `kb.handleModifier()` which is crucial for not letting modifier key state get stuck on MacOS.

So I wrapped pollEvent(), and if we always use the wrapper, handleModifier() will always be called.

### DIFF
--- a/src/tools/winutil.cpp
+++ b/src/tools/winutil.cpp
@@ -1,5 +1,7 @@
 #include "winutil.hpp"
 
+#include "keymods.hpp"
+
 // The default scale should be the largest that the user's screen can fit all three
 // BoE application windows (because they should probably default to match each other).
 double fallback_scale() {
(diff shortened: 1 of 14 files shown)


src/tools/winutil.cpp
```diff
@@ -28,4 +30,22 @@ double fallback_scale() {
...
 	return scale;
+}
+
+// We use many nested event loops in this codebase. Each one of them
+// calls pollEvent() and they each need to remember to call handleModifier()
+// or else modifier keys will claim to be held forever.
+// The best solution for this is to wrap pollEvent() so that it calls
+// handleModifier for us every time.
+bool pollEvent(sf::Window& win, sf::Event& event){
+    if (win.pollEvent(event)){
```
CelticMinstrel, line 41:
```suggestion
    if(win.pollEvent(event)) {
```

PR #494, opened 2024-11-24
Remove/rearrange some settings and improve tiny button/LED handling

### LINKED ISSUE #459: no visual boom for fireball
The boom animation was skipped--damage values didn't show, booms didn't show. Enemies just disappeared, leaving scorch mark and bloodstains.

### LINKED ISSUE #470: Preferences window taller than the main window (and possibly other dialogs?)
With UI scaling in play, having some dialogs be taller than the main game window poses a problem because the player might get important buttons cut off at the bottom of the screen when they thought they found a scale that fit. I know for sure this happens on my screens at 1.5x with the preferences window. The intro dialog is taller than the main window, but not by much, it's probably fine. I can't think of any others.

(Maybe scenarios can make text dialogs bigger than the window? that would be more thorny to fix.)

I think the goal should be for dialogs to be smaller than the main window generally. For preferences, we could achieve this with pagination. There's also room to make minor adjustments in row spacing, and the display alignment widget has extra padding.

Screenshot 2024-11-18 at 11 49 40 AM

NQNStudios:
This removes the 3 settings that appeared to be performance related.

Resetting the help messages is now a button, which triggers a confirmation dialog. Makes more sense than being an LED.

It's no longer possible to set your UI so scale that you can't change it back. The preferences window is also more fitted to the main window now. Fix #470  

Screenshot 2024-11-24 at 5 02 04 PM


CelticMinstrel:
Some thoughts before I look at the changes in detail:

* I kinda aesthetically like it best with the display alignment options at the top. I think putting scale beneath those would still satisfy it being reversible even if you set it too high? Other than if you choose an option which wouldn't physically fit on your screen at all even for the main window, but the correct answer there is to not show those options. I'm not going to say I'm absolutely against leaving it in its current position though.
* Let's move the Never Show Help button down to just above the Reset Help button. Makes more sense to keep the two help options together.
* Is it my imagination or does Reset Help have a smaller font size? It shouldn't.

src/game/boe.main.cpp
```diff
@@ -1296,6 +1295,5 @@ void pause(short length) {
 	// Before pausing, make sure the screen is updated.
 	redraw_screen(REFRESH_NONE);
 	
-	if(get_bool_pref("DrawTerrainFrills", true))
```
CelticMinstrel, line 1299: Removing this should treat it as if the preference is always false. Thus, this becomes "if false then sleep", meaning that the sleep should also be deleted.

src/game/boe.newgraph.cpp
```diff
@@ -124,7 +124,7 @@ void apply_unseen_mask() {
 	rectangle base_rect = {9,9,53,45},to_rect,big_to = {13,13,337,265};
 	bool need_bother = false;
 	
-	if(!get_bool_pref("DrawTerrainFrills", true) || fog_lifted)
+	if(fog_lifted)
```
CelticMinstrel, line 127: This one is… odd. By a strict reading, deleting the preference should follow this path:

* "if not DrawTerrainFrills or fog_lifted then return"
* "if not false or fog_lifted then return"
* "if true or fog_lifted then return"
* "return"

In other words, deleting _the entire function_. But that doesn't sound correct at all, so I suspect this check was wrong to begin with. I wonder if theory is supported by the original source code…?

src/game/boe.newgraph.cpp
```diff
@@ -155,7 +155,7 @@ void apply_light_mask(bool onWindow) {
 	rectangle temp = {0,0,108,84},paint_rect,base_rect = {0,0,36,28};
 	rectangle big_to = {13,13,337,265};
 	bool same_mask = true;
-	if(!get_bool_pref("DrawTerrainFrills", true) || fog_lifted)
+	if(fog_lifted)
```
CelticMinstrel, line 158: This one has the same weird situation as the above one, where a strict reading would suggest deleting the entire function, but that can't be right.

src/game/boe.main.cpp
CelticMinstrel: Oh, I suddenly realized that we want it to always be _true_, not false, so ignore this.

src/game/boe.graphics.cpp
```diff
@@ -1359,7 +1353,7 @@ void boom_space(location where,short mode,short type,short damage,short sound) {
 		return;
 	if((boom_anim_active) && (type != 3))
 		return;
-	if(!get_bool_pref("DrawTerrainFrills", true) && fast_bang)
+	if(fast_bang)
```
CelticMinstrel, line 1356: This looks incorrect. We work on the assumption that the preference is always true, which means the condition becomes "if not true and fast_bang" -> "if false and fast_bang" -> "if false".

So remove the check entirely.

src/game/boe.graphics.cpp
```diff
@@ -1083,7 +1080,7 @@ static void init_trim_mask(std::unique_ptr<sf::Texture>& mask, rectangle src_rec
 //short which_mode;  // 0 top 1 bottom 2 left 3 right 4 up left 5 up right 6 down right 7 down left
 void draw_trim(short q,short r,short which_trim,ter_num_t ground_ter) {
 	/* TODO: Windows has a check for frills being enabled:
-	if(!frills_on || (((current_ground == 2) || (current_ground == 36)) && (PSD[SDF_COMPATIBILITY_FULL_TRIMS] == 0)))
+	if(((current_ground == 2) || (current_ground == 36)) && (PSD[SDF_COMPATIBILITY_FULL_TRIMS] == 0))
```
CelticMinstrel, line 1083: Let's remove the entire TODO note here.

* The PSD at the end is sort of a duplicate of the preference being removed (or might be a slightly related preference that was unique to the Windows version)
* I don't really understand the checks for current_ground, but they seem wrong – hardcoding no frills on two terrains can't be right, just set those terrains to not have frills.

CelticMinstrel:
One other thing I notice in the screenshot: the "Display Alignment" header is not left-aligned with the other headers.

NQNStudios:
>    Is it my imagination or does Reset Help have a smaller font size? It shouldn't.

It's not your imagination. Looks like tiny buttons use point size 9

https://github.com/calref/cboe/blob/df0c46d616904402d1284400db755e0f646a8e18/src/dialogxml/widgets/button.cpp#L58

and LEDs use point size 10.

https://github.com/calref/cboe/blob/df0c46d616904402d1284400db755e0f646a8e18/src/dialogxml/widgets/led.cpp#L32

This is the only tiny button in the game itself. All others are in the scenario editor. Do we bump the text size up a point for tiny buttons across the board and not worry what might happen in the editor?

Or do we add a code path that allows configuring the tiny button text size on a case by case basis?

NQNStudios:
> I kinda aesthetically like it best with the display alignment options at the top.

I tried doing it that way first, and thought it looked really bad. But I forgot to take a screenshot or save a copy of the XML to make it happen. It's kinda tricky doing things back and forth--if I just edit the image to give you an idea of how the other way looks, would you be able to decide from that if you do want it that way?

image

NQNStudios:
I discovered how `led`s and tiny `button`s were all using a hard-coded width for their hitboxes. That was when this PR became more of a sweeping overhaul haha.

src/dialogxml/widgets/button.cpp
```diff
@@ -48,16 +63,16 @@ bool cButton::isScrollable() const {
 	return false;
 }
 
+const int tiny_text_offset = 18;
```
CelticMinstrel, line 66: I'd like this name to be all-caps.

src/dialogxml/widgets/led.cpp
```diff
@@ -86,6 +86,8 @@ bool cLed::manageFormat(eFormat prop, bool set, boost::any* val) {
 	return true;
 }
 
+const int text_offset = 18; // Possibly could be 20
```
CelticMinstrel, line 89: I'd like this to be all-caps… probably `LED_TEXT_OFFSET` would be better too.

src/dialogxml/widgets/button.cpp
```diff
@@ -164,13 +184,19 @@ void cButton::validatePostParse(ticpp::Element& elem, std::string fname, const s
...
-		if(!getText().empty() && !attrs.count("width"))
+		if(type != BTN_TINY && type != BTN_LED && !getText().empty() && !attrs.count("width"))
 			throw xMissingAttr(elem.Value(), "width", elem.Row(), elem.Column(), fname);
 	}
 }
 
 location cButton::getPreferredSize() const {
-	return {btnRects[type][0].width(), btnRects[type][0].height()};
+	int width = btnRects[type][0].width();
+	if(type == BTN_TINY && !getText().empty()){
```
CelticMinstrel, line 194: Hmm… this could probably also be done for `BTN_PUSH` later (the label should probably be centered beneath the button in that case)… though I think it's unused right now so I guess it's not a priority.

src/dialogxml/widgets/button.cpp
```diff
@@ -164,13 +184,19 @@ void cButton::validatePostParse(ticpp::Element& elem, std::string fname, const s
 	if(labelledButtons.count(type)) {
 		if(!attrs.count("color") && !attrs.count("colour") && parent->getBg() == cDialog::BG_DARK)
 			setColour(sf::Color::White);
-		if(!getText().empty() && !attrs.count("width"))
+		if(type != BTN_TINY && type != BTN_LED && !getText().empty() && !attrs.count("width"))
```
CelticMinstrel, line 187: …wait. Why do buttons require width? Other than tiny, LED, and the red push button, isn't the width a fixed function of the type? I'm confused…

rsrc/schemas/dialog.xsd
```diff
@@ -207,6 +207,7 @@
 		</xs:sequence>
 		<xs:attribute name="name" use="required" type="xs:token"/>
 		<xs:attribute name="type" use="required" type="btntype"/>
+		<xs:attribute name="text-size" type="xs:integer"/>
```
CelticMinstrel, line 210: You forgot to add this to the documentation page?

src/dialogxml/widgets/button.hpp
```diff
@@ -74,9 +74,12 @@ class cButton : public cControl {
 	cButton(cDialog& parent,eControlType t);
 private:
 	bool manageFormat(eFormat prop, bool set, boost::any* val) override;
+	void defaultTextSize();
 	std::string fromList;
 	static rectangle btnRects[13][2];
 protected:
+	/// Size of the button's descriptive text
+	short textSize = 0;
```
CelticMinstrel, line 82: I don't think 0 is a good default here.

src/pcedit/pc.menus.mac.mm
```diff
@@ -65,7 +65,7 @@ void init_menubar() {
 	
 	MenuHandler* handler = [[[MenuHandler alloc] init] retain];
 	setMenuCallback([apple_menu itemWithTitle: @"About BoE Character Editor"], handler, @selector(menuChoice:), int(eMenu::ABOUT));
-	setMenuCallback([apple_menu itemWithTitle: @"Preferences…"], handler, @selector(menuChoice:), int(eMenu::PREFS));
+	setMenuCallback([apple_menu itemWithTitle: @"Settings…"], handler, @selector(menuChoice:), int(eMenu::PREFS));
```
CelticMinstrel, line 68: …what are you doing? The menuitem should be Preferences, not Settings.

src/scenedit/scen.menus.mac.mm
```diff
@@ -79,7 +79,7 @@ void init_menubar() {
 	
 	MenuHandler* handler = [[[MenuHandler alloc] init] retain];
 	setMenuCallback([app_menu itemWithTitle: @"About BoE Scenario Editor"], handler, @selector(menuChoice:), int(eMenu::ABOUT));
-	setMenuCallback([app_menu itemWithTitle: @"Preferences…"], handler, @selector(menuChoice:), int(eMenu::PREFS));
+	setMenuCallback([app_menu itemWithTitle: @"Settings…"], handler, @selector(menuChoice:), int(eMenu::PREFS));
```
CelticMinstrel, line 82: …what are you doing? The menuitem should be Preferences, not Settings.

CelticMinstrel:
> if I just edit the image to give you an idea of how the other way looks, would you be able to decide from that if you do want it that way?

I think that screenshot looks good. What do you dislike about it?

I also had another thought about how it could be done:

```
Display Alignment                          Scale
+------------------------------------+  UI      Map
| [ ] Top Left      [ ] Top Right    |  [ ] 1   [ ] 1
|                                    |  [ ] 1.5 [ ] 1.5
|             [ ] Center             |  [ ] 2   [ ] 2
|                                    |  [ ] 3   [ ] 3
| [ ] Bottom Left   [ ] Bottom Right |  [ ] 4   [ ] 4
+------------------------------------+
  [ ] Window
```

src/dialogxml/widgets/button.cpp
CelticMinstrel: Ah, I get it. This is in a block that is only reached if the button is one of those three types. So, removing the check entirely is probably correct after all (and also implement some kind of automatic size detection for `BTN_PUSH`).

src/dialogxml/widgets/button.hpp
NQNStudios: It doesn't stay 0, the 0 indicates that defaultTextSize() should do its thing.

src/pcedit/pc.menus.mac.mm
NQNStudios: The actual menu item that's in there is "Settings..." on Mac for both cases for me. Which is very confusing because I see in the `xib` files it should be "Preferences..."

But at actual runtime, my scenario editor and character editor menu items say Settings, not Preferences, and don't work (are disabled) unless these strings match.

src/scenedit/scen.menus.mac.mm
NQNStudios: duplicate comment

NQNStudios:
Ok I'll flip the scale LEDs and the display alignment widget tomorrow.

src/dialogxml/widgets/button.hpp
```diff
@@ -74,9 +74,12 @@ class cButton : public cControl {
 	cButton(cDialog& parent,eControlType t);
 private:
 	bool manageFormat(eFormat prop, bool set, boost::any* val) override;
+	void defaultTextSize();
```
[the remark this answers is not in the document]
NQNStudios: I'm not sure how that would work.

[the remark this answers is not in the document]
CelticMinstrel: I was thinking it would take the button type as argument and return the default font size.

src/dialogxml/widgets/button.hpp
CelticMinstrel: I'm not sure if that actually makes sense though…? It would make sense if defaultTextSize ran _after_ parsing, but that's not what I see here.

src/pcedit/pc.menus.mac.mm
CelticMinstrel: That's _really_ weird. That's not even the case for me – when I run the scenario editor, I see Preferences, not Settings.

src/dialogxml/widgets/led.cpp
CelticMinstrel: Uhh…? You changed the name here but it looks like you missed the use site?

CelticMinstrel: Ah, fixed in later commit.

src/pcedit/pc.menus.mac.mm
NQNStudios: Could this be a scons vs xcode problem? Is the xcode build doing something with the xib file that scons isn't?

NQNStudios: I see in the scons build output:

> ibtool --compile build/obj/menus/scenedit.nib rsrc/menus/scenedit.xib
Install file: "build/obj/menus/scenedit.nib" as "build/Blades of Exile/BoE Scenario Editor.app/Contents/Resources/scenedit.nib"

So I should look at that nib file.

NQNStudios: It's not human-readable and Xcode isn't showing me the contents either.

src/dialogxml/widgets/button.hpp
NQNStudios: Oh! I misunderstood the call order. Yeah, it doesn't need default 0.

src/pcedit/pc.menus.mac.mm
NQNStudios: I'll revert the change and open an issue for it.

NQNStudios:
Screenshot 2024-11-26 at 12 15 28 PM

I think this is pretty dang good. And you can see it fits inside the main window.

NQNStudios:
@CelticMinstrel two commits came along for the ride here to fix some things that bothered me while testing the rest of it with a replay. They *could* split into another PR, but if there's any more back-and-forth to do with the rest, it'd be nice to keep at least the first one in this branch until we lock in the rest. Also, I think both commits are good changes with low risk and might as well not be split out.

[aff33dd](https://github.com/calref/cboe/pull/494/commits/aff33ddc50dff8208ad19646bbb5833e3fc77369) disables the Spiderweb Software splash screen when running a replay with the No Splash Screen setting (the setting usually only disables the 2nd splash screen.) This saves a lot of time when running replays repeatedly.

[166d71a](https://github.com/calref/cboe/pull/494/commits/166d71a4ffae1f6f0061390c3ea92e122112e818) Since I implemented replaying the startup menu buttons, they've had that visual bug where the rest of the screen is rendered black. I fixed that.

If you like the layout I shared just above, this is ready to merge.

NQNStudios:
Are these blue ones BTN_PUSH?

Screenshot 2024-11-27 at 3 28 56 PM

I didn't know which ones BTN_PUSH was, so I wasn't prioritizing adding auto-width to them. But I see now they're right there in the main scenario editor screen. And should be easy enough to add in.

rsrc/schemas/dialog.xsd
```diff
@@ -229,6 +236,7 @@
 				<xs:attributeGroup ref="rect-size"/>
 				<xs:attributeGroup ref="position"/>
 				<xs:attributeGroup ref="font"/>
+				<xs:attribute name="text-side" default="right" type="textSide"/>
```
CelticMinstrel, line 239: I don't really like this name for it. "alignment" might be a bit better but is also not that great. Maybe "attachment"? Not sure I like that name either… got any other ideas?

CelticMinstrel:
I like the layout in the latest screenshot. Putting the scales extremely close together like that was a really good idea too (normally LEDs would be spaced out a bit more).

NQNStudios:
@CelticMinstrel How about `label-pos`?

rsrc/schemas/dialog.xsd
```diff
@@ -63,6 +63,12 @@
 			<xs:enumeration value="led"/>
 		</xs:restriction>
 	</xs:simpleType>
+	<xs:simpleType name="textSide">
```
CelticMinstrel, line 66: Looks like you missed this case in the rename.

NQNStudios:
@CelticMinstrel should finally be ready to go in. For that conflict, is it easy for you to resolve? The file should just be deleted.

CelticMinstrel:
It means I can't merge it just by clicking a button. It probably wouldn't be that difficult to do on the command-line, not quite sure when I'd get to that tho.

NQNStudios:
Maybe there's an easy way for me to fix it without a full rebase.

PR #498, opened 2024-11-26
Refactor boom code & add guaranteed central explosion for fireballs

NQNStudios:


CelticMinstrel:
It's not a bad idea.

NQNStudios:
This affects both Fireball and Firestorm, which both use the same explosion, leave ashes, and without this change will be silent/have no visual feedback if you don't hit anyone.

PR #502, opened 2024-11-28
Some scenedit improvements, replay test for issue #8

NQNStudios:
* Improved an error message from the scenario editor that boggled my mind while I was trying to make a test scenario for issue #8 
* Added a preprocessor definition `DEBUG` for scons builds
* Created 2 test scenarios with 1 replay test each for issue #8 - one shows a new format scenario, another a legacy scenario
* Debug builds run from the build directory will now find scenarios in the `test/replays/scenarios` folder.
    * You might not love the idea of supporting multiple scenario folders at the same time, but the alternative here is to have the replay test script `test/replays.sh` *copy* test scenarios into the user's scenario dir, which is infinitely worse
        * Well I guess another option would be to keep test scenarios in the same folder as valleydy, etc., but that feels wrong? And I'm not sure that that folder gets recursively searched anyways?
    * I am planning to use this feature for my own scenario packaging system for the itch.io edition, so it would be nice to have it upstream


NQNStudios:
I now need to rebase this because #501 changed how we need to check if the scons build is debug or not

NQNStudios:
And while I'm at it I will remove the constraint in the scenario editor that prompted opening #503

NQNStudios:
Ok, rebased and removed the outdoor node constraint here.

PR #474, opened 2024-11-18
Use macos-13 for ci

NQNStudios:
This makes the required jump between macos-12 and macos-13, and fixes resulting CI errors that I found. Fix #473

PR #505, opened 2024-11-30
Use macos-13 for ci

NQNStudios:
I had opened #474 from my master branch, which was starting to get awkward. This is the same change.


NQNStudios:
@CelticMinstrel It looks like they've permanently disabled macos-12 now. Probably time to merge this.

PR #519, opened 2024-12-01
Sounds

NQNStudios:
* Changed shop purchase sound effects where Yum! definitely didn't fit. For alchemy, mage spells, and priest spells, I made the sound the same as the sound when you perform those actions. I figure this actually makes sense because the teacher would show you how to do the thing.
* For skill training, I made the sound be the same as leveling up. I think this makes sense.
* When something is too heavy to pick up, the "Argh!" labeled as 'Action Failure' in the editor sound picker makes a lot of sense--you tried to pick it up and were frustrated.
* I changed beep() so every platform just plays sound effect 1, the one labeled "low beep" [here](http://openboe.com/docs/editor/appendix/Sounds.html) that's labeled "Cancel target lock" in the editor picker. I think this reasonable.
* Removed `ding()` which was an unused alias for `beep()`

### DIFF
--- a/src/game/boe.dlgutil.cpp
+++ b/src/game/boe.dlgutil.cpp
@@ -293,8 +293,7 @@ void handle_sale(int i) {
 			else if(!take_gold(cost,false))
 				ASB("Not enough gold.");
 			else {
-				// TODO: This seems like the wrong sound
-				play_sound(62);
+				play_sound(8);
 				ASB("You buy an alchemical recipe.");
 				univ.party.alchemy[base_item.item_level] = true;
 			}
@@ -352,8 +351,7 @@ void handle_sale(int i) {
 			else if(!take_gold(cost,false))
 				ASB("Not enough gold.");
 			else {
-				// TODO: This seems like the wrong sound
-				play_sound(62);
+				play_sound(25);
 				ASB("You buy a spell.");
 				univ.current_pc().mage_spells[base_item.item_level] = true;
 				give_help(41,0);
@@ -370,8 +368,7 @@ void handle_sale(int i) {
 			else if(!take_gold(cost,false))
 				ASB("Not enough gold.");
 			else {
-				// TODO: This seems like the wrong sound
-				play_sound(62);
+				play_sound(24);
 				ASB("You buy a spell.");
 				univ.current_pc().priest_spells[base_item.item_level] = true;
 				give_help(41,0);
@@ -401,8 +398,7 @@ void handle_sale(int i) {
 			else if(!take_gold(cost, false))
 				ASB("Not enough gold.");
 			else {
-				// TODO: No idea what's a good sound here.
-				play_sound(62);
+				play_sound(7);
 				ASB("You learn a little...");
 				active_shop.takeOne(i);
 				univ.current_pc().skills[skill]++;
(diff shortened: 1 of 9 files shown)


CelticMinstrel:
Okay, so, I don't think beep() should be made to play a game sound. I think it's used in places like error dialogs where the system alert is quite appropriate (though it's unfortunate that Linux doesn't implement it).

Besides the cases where you removed a TODO comment, are there a lot of calls to beep()?

NQNStudios:
There are a fair amount of beep calls in all three programs.

```
$ rg "beep\(\);"
src/scenedit/scen.townout.cpp
1238:		if(cur_loc.x == 0) beep();
1241:		if(cur_loc.x >= scenario.outdoors.width() - 1) beep();
1244:		if(cur_loc.y == 0) beep();
1247:		if(cur_loc.y >= scenario.outdoors.height() - 1) beep();
1432:			beep();
1438:			beep();

src/pcedit/pc.editors.cpp
383:			beep();
393:			beep();
412:			} else beep();
426:				else beep();
447:			} else beep();
461:				else beep();
495:				} else beep();
509:					else beep();

src/tools/winutil.hpp
56:void beep();

src/scenedit/scen.core.cpp
3437:			beep();
3456:			beep();
3604:		beep();
3619:			beep();

src/game/boe.town.cpp
1260:				beep();

src/game/boe.dlgutil.cpp
345:				beep();
362:				beep();
391:				beep();
410:		beep();
786:				beep();

src/game/boe.party.cpp
287:		beep();
293:		beep();
321:		beep();
```

CelticMinstrel:
That does seem like a lot, though it's hard to judge them with no context at all… how many of them are "external" events not related to game logic?

The system alert is not just a sound. It is a generic "alert the user" action. If the speakers are muted, for example, on MacOS it will flash the menubar to draw your attention. So, replacing it with a sound does not create an equivalent effect. There may be an argument that nothing in Blades of Exile merits a true system alert. If that's the case, beep() should be removed altogether and just use play_sound(). But I think there are probably at least a few places that merit a system alert.

NQNStudios:
- [x] the beep in boe.items.cpp is replaced with a real sound in this PR
- [ ] All the beeps in boe.dlgutil.cpp() are for scenario errors that also call showError(), except for one when you click Record and you can't.
- [ ] the beep in boe.town.cpp also looks like it's for an error that's not supposed to happen
- [ ] The 3 beeps in boe.party.cpp are for internal errors that aren't supposed to happen.

I agree now that we should keep beep() the same. On linux, it can be `play_sound(1)` which is better than nothing. And for the one ingame beep that could be normal, I think `play_sound(1)` would be better although we should be just hiding the Record button instead.

CelticMinstrel:
> On linux, it can be play_sound(1) which is better than nothing.

The closest I know of to a "correct" answer on Linux is `std::cout << '\a';`. However, I think that only works if stdout is pointing to a TTY, so it's not sufficiently generic to actually use as the Linux implementation. I don't know if there exists a more generic method on Linux.

NQNStudios:
> All the TODO comments re-added by the revert could also be changed to play_sound(1), couldn't they? Though I don't know if there's any better sound to use in those cases.

@CelticMinstrel I don't think there's a better sound, since they're in the PC editor so gamey sound effects wouldn't be expected. I've made them `play_sound(1)`.

I could make beep() print the bell character on linux, but I can't find the hard drive with my linux environment on it so I wouldn't be able to test.

CelticMinstrel:
> I could make beep() print the bell character on linux, but I can't find the hard drive with my linux environment on it so I wouldn't be able to test.

I'd prefer this, but I'm not going to consider it a blocker. If someone else knows how to do it in a context where there may not even _be_ a tty, it can be done later.

Issue #233, opened 2020-01-17
Buttons in inventory window visible during shopping


x-qq:
![zh0V](https://user-images.githubusercontent.com/55463047/72582951-248f5800-38ed-11ea-85c0-70959250bc46.png)

The Use, Give, Drop buttons were not visible in the original game.

wrldwzrd89:
This looks like a simplification gone awry - the reason the Use/Give/Drop buttons weren't orignally displayed while shopping is because they don't make sense to use in that context.

CelticMinstrel:
Right, and they _still_ don't make sense in that context, so they still shouldn't be shown.

PR #520, opened 2024-12-02
Simplify scary logic around inventory item buttons.

NQNStudios:
The logic that was causing #233 was marked as kludgy and in need of a redo. I've done that, and I think the new logic is more correct.

### DIFF
--- a/src/game/boe.consts.hpp
+++ b/src/game/boe.consts.hpp
@@ -120,9 +120,6 @@ enum eItemButton {
 	MAX_eItemButton // keep last
 };
 
-const int ITEMBTN_ALL = 10; // use, give, drop, info
-const int ITEMBTN_NORM = 11; // give, drop, info
-
 // PC buttons
 enum ePlayerButton {
 	PCBTN_NAME = 0,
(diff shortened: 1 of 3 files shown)


CelticMinstrel:
1. Does this render `ITEMBTN_NORM` and `ITEMBTN_ALL` as entirely unused? If so, they can be deleted.
2. This doesn't break the ID / Ench / etc buttons, does it?

NQNStudios:
I don't think so, but it needs testing. Incidentally, how would I make a party with an unidentified item? Edit a scenario and have them pick it up? Does the PC editor allow adding items? Should I just play the game until I find some? I do need to keep just testing VoDT so maybe that's what I'll do.

CelticMinstrel:
The PC editor and the in-game debug mode both support adding items, but if I recall correctly, they automatically identify the items added. Probably the best way to get unidentified items is to enter a scenario and steal some stuff. (Picking stuff up off the ground is fine too, but those kinds of items often have the "Always Identified" flag.)

NQNStudios:
- [x] test can use items in combat, but not while targeting
- [ ] test doesn't break ID/Enchant/etc buttons
- [ ] see about removing `ITEMBTN_NORM` and `ITEMBTN_ALL`

CelticMinstrel:
Just a note: the full list of buttons for point 2 is identify (ID), sell, enchant (Ench), and recharge (Inc).

NQNStudios:
@CelticMinstrel I tested all those cases, I think it's good to go.

PR #521, opened 2024-12-02
Cancel item target

NQNStudios:
I forgot to have Escape cancel item target mode.

While I'm at it, I want to make identification cancel in advance if there are no unidentified items, and it should end after identifying the last item (as was TODO noted).

### DIFF
--- a/src/game/boe.party.cpp
+++ b/src/game/boe.party.cpp
@@ -605,11 +605,9 @@ void do_mage_spell(short pc_num,eSpell spell_num,bool freebie) {
 				if(all_identified){
 					sstr << "already ";
 				}else{
-					for(cPlayer& pc : univ.party){
+					for(cPlayer& pc : univ.party)
 						for(cItem& item : pc.items)
 							item.ident = true;
-						pc.combine_things();
-					}
 				}
 				sstr << "identified.";
 				ASB(sstr.str());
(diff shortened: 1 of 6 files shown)


CelticMinstrel:
> always combine_things() after identifying a held item 

What does this do, again? It seems possible wrong but I'm not sure.

src/game/boe.party.cpp
```diff
@@ -2604,3 +2620,12 @@ short party_size(bool only_living) {
...
 }
+
+bool all_items_identified() {
+	bool all_identified = true;
+	for(cPlayer& pc : univ.party)
+		for(cItem& item : pc.items)
+			if (item.variety != eItemType::NO_ITEM)
+				all_identified &= item.ident;
+	return all_identified;
+}
```
CelticMinstrel, line 2631: Hmm… I wonder, should this maybe be a member function of `cParty` instead?

NQNStudios:
> > always combine_things() after identifying a held item
> 
> What does this do, again? It seems possible wrong but I'm not sure.

It combines things into stacks. So if you identify a stack of items that can be combined with a stack of items that you already identified, the two combine in one slot.

CelticMinstrel:
> It combines things into stacks. So if you identify a stack of items that can be combined with a stack of items that you already identified, the two combine in one slot.

If this happens while I'm in the middle of identifying things, my reaction will probably be "Huh? Where'd it go?". So I think it's probably not good.

NQNStudios:
@CelticMinstrel that's a good point, so I reverted that part.

PR #527, opened 2025-01-09
Fix led labels

### LINKED ISSUE #297: LEDs should generally contain their label text
The dialog engine in the original game didn't have any association between an LED button and its label text. The LED button was just a button, and the label text was just an unrelated static text.

Associating them however adds the nice feature that you can click the label text to activate the button. So, I'd like to update as many dialogs as possible to do this.

#274 and #273 notes a number of dialogs that could use the treatment. There may be more, and some of those might not merit it.

NQNStudios:
I've gone through all the files that my grep pattern identified as having textless `led` elements (#297). This should fix all of them, and close #297.

That is unless `led`s are created programatically anywhere in the codebase. But if they all have to be declared in xml then I think I can guarantee I changed them all.

### DIFF
--- a/rsrc/dialogs/edit-item-shortcut.xml
+++ b/rsrc/dialogs/edit-item-shortcut.xml
@@ -46,8 +46,7 @@
 	<text top='154' left='159' width='119' height='26'>Chance of placing: (0-100%)</text>
 	<text top='154' left='286' width='99' height='36'>Number of item to place:</text>
 	<text top='154' left='395' width='119' height='26'>Chance of placing: (0-100%)</text>
-	<led name='owned' state='off' top='138' left='215'/>
-	<text top='134' left=' 50' width='155' height='14'>Items are always property:</text>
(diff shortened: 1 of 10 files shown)


CelticMinstrel:
So, this means that places where the LED was to the right of its label have moved the LED to the left instead, right?

Did you verify that the text size of the labels didn't change?

NQNStudios:
Yeah all of these LEDs move to the left of their text. I verified that
everything looks good, although I didn’t compare text size directly between
old and new.

CelticMinstrel:
Is the default font size the same for labels and LEDs? I don't remember.

NQNStudios:
Here is before
image

Here is after
image

Text size looks the same to me. We could get fiddly about having the two groups of 3 leds horizontally align together, but the dialogs I edited for this PR are all in the scenario editor so that level of aesthetic attention doesn't feel warranted.

PR #531, opened 2025-01-12
Fix win-scons CI

### LINKED ISSUE #514: win-scons ci is failing consistently

NQNStudios:
When finished this will close #514

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,119 +11,119 @@
 	},
 
 	jobs: {
-		macos-xcode: {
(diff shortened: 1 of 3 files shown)


NQNStudios:
Thanks to the right prodding from @CelticMinstrel, I got this to work again.

I changed SConstruct so when the compiler check fails, it runs our own test compilation, the output of which will actually be printed in the log. We could take that out now that it's fixed, but I think we might as well leave it. The nebulous "something wrong with your compiler" error is my worst nightmare and we've seen it before.

Issue #417, opened 2024-08-28
some signs in Fort Talrus showing blank text (mac)


NQNStudios:

CelticMinstrel:
Noting here that the issue doesn't appear to be in the scenario itself. I looked at the scenario data, and I can see the two missing signs in Fort Talrus. However, the scenario editor _also_ says they're blank.

NQNStudios:
That’s weird — when I checked in the scenario editor I saw all the proper
text for signs I checked

CelticMinstrel:
I don't know if this is relevant, but I was loading the unpacked version of Valley of Dying things that's committed to the repository.

NQNStudios:
I can't remember which file I opened to check--could have been the unpacked source, could have been the one in the app package.

If you're looking into fixing this one, maybe we should start assigning the issues to keep clear on who's tackling what.

NQNStudios:
This was an instance of the font corruption bug we fixed.

PR #534, opened 2025-01-16
Fixing text buffer texture/font corruption (#479)

### LINKED ISSUE #479: Text invisible in the message log.
I don't know when this happened, but I was testing VoDT and I noticed the message log had become blank except for a single question mark.

Screenshot 2024-11-21 at 12 54 33 PM

### LINKED ISSUE #532: Replays from older builds cause crash
34dd31cf51f5307025c9e08074483fdc038c7040 removed 3 preferences. Replays created with builds prior to 34dd31cf51f5307025c9e08074483fdc038c7040 have 3 extra preferences in their recorded `load_prefs`, and these cause a crash in new builds.

The quick-and-dirty solution to make those replays work, is to delete those 3 lines in the action log xml, as I have done for #479.

But it would be more preferable if the replay system could skip obsolete preferences in a `load_prefs`, and vice-versa, I should make sure that *new* preferences can be missing from a `load_prefs` element without causing a crash. (They'd be treated as the default in that case.)

### LINKED ISSUE #533: Replays from older builds click the wrong startup button
https://github.com/calref/cboe/pull/500 rearranged and combined startup menu buttons.

Replays from builds prior to https://github.com/calref/cboe/pull/500 still encode startup button indices according to the old layout, so they don't work in newer builds if they click a startup button that moved.

This could be hairy to implement a real fix for -- (How can you programmatically determine if the build comes after a specific commit SHA?). Another solution I'd be okay with, is just directly editing the indices in all the logs we have in `test/replays`. (And then we promise not to rearrange buttons again.)

NQNStudios:
Debugging #479 has become a meandering process, with some other bugs needing to be fixed along the way. Might as well keep them in 1 PR.

* #479 demonstrates that the contents of the text buffer are NOT irrelevant for reproducing bugs. So I set up recording/replay for the burma shave easter egg. This also makes an easy way to mess with the buffer state when debugging (just mash &/\*/&/\*/&/\* n times)
* When a replay throws an error, it puts up a showError() dialog. If the next action is a control_click, the system will try to click that control on the error dialog--which is totally divergent from the replay's intended behavior. So we should just stop replaying when an error happens.
* If you have a long replay and want to run it very fast, but then slow down when you get to the sequence that reproduces your bug, now you can add a `<change_fps>` to your replay to achieve that.
* Fixes for the 2 legacy replay errors that I opened yesterday

I still am quite stumped on the real error for #479. I asked for help in the SFML Discord channel, here's a [link to that thread](https://discord.com/channels/175298431294636032/175298431294636032/1329466205366390825) (you'll need a discord account and to join the SFML group to see it). But anyway, this link makes it easy for me to consult back to what people suggested.

EDIT: Changed the title to clarify one thing I do know about this bug: the text buffer array still contains the correct text. This is a case of the `sf::Text` that's rendered onto text_area_gworld being corrupt somehow. Maybe the font gets corrupted.

Fix #479 
Fix #532 
Fix #533


NQNStudios:
@CelticMinstrel This bug was caused by this line:

https://github.com/calref/cboe/pull/534/commits/62ddf3d0b3f1baf7b2b9330d732832957f7b1c5b#diff-0352715fc63f00d5823113f53777233ac00d70eb82df14dc28b2e9366710ee86L961

When creating a new PC, the empty slot was being filled by this cPlayer constructor which set the player name to "\n". Then, pick_pc_name() would be called, putting the "\n" in as the text field's default value.

Now, I don't know why, but it appears that the larger issue here is that if we accidentally try to render a newline character, the font object becomes borked. This probably explains #454 (one instance of which this PR also fixes) and maybe #417 as well. We need to find all the ways we might end up trying to do that... And maybe add an assertion that there are no newlines in strings being drawn.

In this case, setting the default player name to empty string, fixes #479 as reproduced by the log in that thread.

Was there any reason to set the player's default name to a newline and not an empty string?

NQNStudios:
According to the blame, it looks like the "\n" default name has been in place since 16 years ago when Mac and Windows code were separate.

src/game/boe.actions.cpp
```diff
@@ -3743,11 +3757,11 @@ bool is_sign(ter_num_t ter) {
 	return false;
 }
 
-bool check_for_interrupt(){
+bool check_for_interrupt(std::string confirm_dialog){
 	using kb = sf::Keyboard;
 	bool interrupt = false;
 	sf::Event evt;
-	if(replaying && has_next_action() && next_action_type() == "handle_interrupt"){
+	if(replaying && confirm_dialog == "confirm-interrupt-special" && has_next_action() && next_action_type() == "handle_interrupt"){
```
CelticMinstrel, line 3764: It looks like you made up a new dialog name that doesn't exist here – the old code was using "confirm-interrupt" but you changed it to "confirm-interrupt-special". Same thing in the header file for the default parameter value.

CelticMinstrel:
> According to the blame, it looks like the "\n" default name has been in place since 16 years ago when Mac and Windows code were separate. I'm thinking its purpose is obsolete and an empty string is fine?

I found the `"\n"` in the original source code released by Spiderweb Software - at party.c:528. I have no idea why it was there, but it used to be an `sprintf` call. If you used blame I'm guessing you saw this as well.

> Now, I don't know why, but it appears that the larger issue here is that if we accidentally try to render a newline character, the font object becomes borked.

This strikes me as kind of weird. Are we doing something strange in our rendering? I don't think this happens in other projects where I use SFML – rendering a newline should just be a no-op. Though admittedly, other projects probably only render a newline as part of a run of text and not alone… or it's possible I had logic to split on newline so the newline actually isn't being rendered.

NQNStudios:
I’ll dig into the nitty gritty of why it happens—we should probably know
*exactly* why.

NQNStudios:
I've modified draw_startup() to just directly do some tests.
```
void draw_startup(short but_type) {
	
        // We can change this line to demonstrate what breaks the normal text declared below
        sf::Text text("\n", *ResMgr::fonts.get("plain"), 12);


	mainPtr.draw(text);
	sf::Text text2("This text should appear normal", *ResMgr::fonts.get("plain"), 12);
	text2.setPosition(100, 100);
	mainPtr.draw(text2);
}
```

This first version, just putting "\n" in an sf::Text, doesn't break anything.
Screenshot 2025-01-18 at 1 33 28 PM

A newline in the middle of a string also doesn't break anything. Actually, SFML handles the newline.

`sf::Text text("text\nwith a newline in it", *ResMgr::fonts.get("plain"), 12);`

Screenshot 2025-01-18 at 1 34 42 PM

Now, using our cboe library functions:
```
	rectangle dest = { 0, 0, 50, 400};
	TextStyle style;	
	win_draw_string(mainPtr,dest,"\n",eTextMode::LEFT_TOP,style);
```
Screenshot 2025-01-18 at 1 41 33 PM

Still good.

win_draw_string with a newline in the middle:
Screenshot 2025-01-18 at 1 43 12 PM

Now I'm getting surprised this all still works.

#479 was caused by the newline being in the string for a cTextField, and those use draw_string_hilite so I tried that next:

```
	draw_string_sel(mainPtr,dest,"\n",style,{{0, 1}}, Colours::BLUE);
```
Screenshot 2025-01-18 at 1 50 13 PM

And still not reproducing the bug.

Still getting to the bottom of this.

NQNStudios:
No more messing around, I tried just launching a text field with "\n" in it:

```
	static bool ran_dialog = false;
	if(!ran_dialog){
		cDialog pcPickName(*ResMgr::dialogs.get("pick-pc-name"), nullptr);
		pcPickName["name"].setText("\n");
		pcPickName["okay"].attachClickHandler([](cDialog& me, std::string _1, eKeyMod _2) -> bool {
			me.toast(true);
			return true;
		});
		
		pcPickName.run();
		ran_dialog = true;
	}

```

But this still didn't break anything!

NQNStudios:
I still honestly don't know how to reproduce this bug without running the very long replay that entails printing to the buffer hundreds of times and doing all kinds of weird state changes.

NQNStudios:
Could the newline in the default name have been put there to avoid a segfault/out-of-bounds here?

https://github.com/calref/cboe/blob/3daf1d55e42abb4f9118f98f03a2de1cebda1b5f/src/game/boe.party.cpp#L2225

CelticMinstrel:
> Could the newline in the default name have been put there to avoid a segfault/out-of-bounds here?

Seems plausible, but that wouldn't be a segfault even on an empty string, since strings need to be null-terminated (even `std::string` needs that so that `c_str()` works), and also because a short string doesn't allocate any memory (there's no heap pointer to dereference). It would probably be an out-of-bounds assertion in some debug C++ runtimes though.

PR #536, opened 2025-01-18
Quality of life: Spellcasting

### LINKED ISSUE #535: load party, press P: "Priest spell Light not implemented for town mode."
Just a wonky error message. It should say something like "Cannot recast: need to cast a spell first."

NQNStudios:
This makes changes to the spellcasting UI.

* M or P to recast will no longer default to Light or Minor Bless/Minor Heal. You need to cast something before recast becomes available. This fixes #535 and I think it's disorienting when I've just started the game and M casts Light in a town that's fully lit, so the change is generally good I'd say.
* I implemented a recasting hint in the text bar, which was one of the things I mentioned in my quality-of-life checklist https://github.com/NQNStudios/cboe/issues/16. It should move status icons to the left when necessary, to fall between the left text and right text. I think with very long spell names, this could overlap with the text, which would be bad. But I don't know what statuses go in the text bar so I haven't tested/handled that case yet.
Screenshot 2025-01-18 at 5 39 31 PM

* Sometimes when my eyes glaze over, I think I'm casting the spell on the wrong side of the LED. I thought there was a bug when I cast Long Light instead of Dumbfound (even though I know the distance between the two is pretty large -- I wasn't paying much attention). I thought it would be nice to highlight the name of the selected spell. Light green seemed to make more sense than red for that, because the LED turns green. Then I made the caster/target selection texts also use light green instead of red, to match. Uncastable spells are grey.

Screenshot 2025-01-18 at 5 39 15 PM

Honestly still haven't decided if I like how it looks, so feedback is welcome.

### DIFF
--- a/rsrc/dialogs/cast-spell.xml
+++ b/rsrc/dialogs/cast-spell.xml
@@ -28,12 +28,12 @@
 	<button name='target5' type='small' def-key='shift 5' relative='pos-in pos-in' rel-anchor='prev' top='24' left='0'><key/></button>
 	<button name='target6' type='small' def-key='shift 6' relative='pos-in pos-in' rel-anchor='prev' top='24' left='0'><key/></button>
 	<!-- To the right of target buttons, green arrows to highlight choice -->
-	<text name='arrow1' relative='neg pos-in' anchor='hp1' top='0' left='20' width='32' height='16' colour='light_green'>-&gt;</text>
-	<text name='arrow2' relative='pos-in pos-in' rel-anchor='prev' top='24' left='0' width='32' height='16' colour='light_green'>-&gt;</text>
-	<text name='arrow3' relative='pos-in pos-in' rel-anchor='prev' top='24' left='0' width='32' height='16' colour='light_green'>-&gt;</text>
-	<text name='arrow4' relative='pos-in pos-in' rel-anchor='prev' top='24' left='0' width='32' height='16' colour='light_green'>-&gt;</text>
-	<text name='arrow5' relative='pos-in pos-in' rel-anchor='prev' top='24' left='0' width='32' height='16' colour='light_green'>-&gt;</text>
-	<text name='arrow6' relative='pos-in pos-in' rel-anchor='prev' top='24' left='0' width='32' height='16' colour='light_green'>-&gt;</text>
+	<text name='arrow1' relative='neg pos-in' anchor='hp1' top='0' left='20' width='32' height='16' colour='light-green'>-&gt;</text>
+	<text name='arrow2' relative='pos-in pos-in' rel-anchor='prev' top='24' left='0' width='32' height='16' colour='light-green'>-&gt;</text>
+	<text name='arrow3' relative='pos-in pos-in' rel-anchor='prev' top='24' left='0' width='32' height='16' colour='light-green'>-&gt;</text>
+	<text name='arrow4' relative='pos-in pos-in' rel-anchor='prev' top='24' left='0' width='32' height='16' colour='light-green'>-&gt;</text>
+	<text name='arrow5' relative='pos-in pos-in' rel-anchor='prev' top='24' left='0' width='32' height='16' colour='light-green'>-&gt;</text>
+	<text name='arrow6' relative='pos-in pos-in' rel-anchor='prev' top='24' left='0' width='32' height='16' colour='light-green'>-&gt;</text>
 	<!-- To the right of target buttons, PC HP -->
 	<text name='hp-head' size='large' relative='pos-in pos-in' anchor='target-head' top='0' left='55' width='223' height='16'>HP:</text>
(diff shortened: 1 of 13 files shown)


NQNStudios:
> * It should move status icons to the left when necessary, to fall between the left text and right text. I think with very long spell names, this could overlap with the text, which would be bad. But I don't know what statuses go in the text bar so I haven't tested/handled that case yet.

The thing to do about this would probably be to calculate how much space is taken up by the name, APs, and active status icons, then squish the spell hint horizontally as needed.

CelticMinstrel:
Hmm… my first reaction to the spell hint was "that sounds bad, there just isn't enough space", and I still mostly feel that way (what about characters that can cast _both_ mage _and_ priest spells)? However, if it's only shown as part of the combat AP indicator, then I think I can live with it. In that case we can drop the status icons altogether – the only statuses shown there are whole-party statuses (fly, detect life, firewalk, stealth), most of which have no relevance in combat. (I think only firewalk could have combat relevance?)

So, to be clear: in combat we can show this indicator, and outside of combat we can show the status icons, but not both at the same time. Does that work? I'm not sure if it undermines the intent of your hint.

NQNStudios:
Here's a redesign of the spell window that aligns the numbers, buttons, statuses etc. and colorizes HP/SP numbers

Screenshot 2025-01-20 at 1 22 45 PM

NQNStudios:
> So, to be clear: in combat we can show this indicator, and outside of combat we can show the status icons, but not both at the same time. Does that work? I'm not sure if it undermines the intent of your hint.

I think this is just fine.

> (what about characters that can cast both mage and priest spells)?

I've just made it so attempting to recast the opposite spell type of the hint, toggles the hint to the other spell type, so you get to see it before repeating the input to do the recast.

I made this part of the spellcasting action that we already record and replay, so it should work in replays.

CelticMinstrel:
Regarding the colours, I think my logic behind choosing those colours was that they are the basic HTML colour set… or perhaps it was that it's the basic Win32 console colour set, or the union of the two. So I think I'd prefer to keep all the colours that were defined before, though moving them to a colour map definitely makes the code cleaner.

One other note on the colours – for two-word colours, could we replace the underscores with hyphens? So for example `color="title-blue"` instead of `color="title_blue"`.

Also, I think you missed updating the schema for the colours change? Unless I missed it.

-----

I notice you added a documentation comment in the cpp (asPeaceful). Wouldn't that make more sense in the header?

NQNStudios:
> Also, I think you missed updating the schema for the colours change? Unless I missed it.

`color`/`colour` actually doesn't have its acceptable values specified in the schema.

NQNStudios:
> I notice you added a documentation comment in the cpp (asPeaceful). Wouldn't that make more sense in the header?

I think it would actually make the most sense right next to the declarations of all the spells. I added that note in because I thought all the instances of `when(WHEN_COMBAT).asPeaceful()` were bugs and they're not.

CelticMinstrel:
> For the ones that exist as `Colours` constants can I use those values instead of the ones that were defined before? It'd be really weird/inconvenient if the xml couldn't make controls match colors with the programmatic values

How different were they? If they're reasonably similar colours then the version in `Colours` should be fine.

> `color`/`colour` actually doesn't have its acceptable values specified in the schema.

Hmm, maybe that should be added, but if it's already not there then it won't block this PR.

NQNStudios:
> How different were they? If they're reasonably similar colours then the version in Colours should be fine.

They're pretty similar, and also nothing was referencing the variants that would have come from the old `parseColor()` so there's no functional change.

> Also, you misspelled "silver".

That I did. Funny that it wasn't fuchsia that tripped me up!

CelticMinstrel:
Okay, I did a comparison of the colours. The ones that are different are:

* red and blue use 0xdd instead of 0xff
* yellow adds 0x31 blue to mellow it out a bit
* navy is even darker - 0x39 instead of 0x80
* teal is a significantly different colour

The differences mostly aren't noticeable unless you look closely (except for teal). I did wonder if perhaps some of the "extra" colours that use 0xff should instead either use 0xdd or fill the gap with 0x31, so I tried this out:

* lime with 0xdd is noticeably darker but still quite bright
* aqua as 00dddd is noticeably darker but still quite bright
* aqua as 00ffdd is about the same as 00ffff while 00ddff is about the same as 00dddd
* fuchsia as dd00dd is noticeably darker but still quite bright
* fuchsia as ff00dd is about the same as ff00ff while dd00ff is about the same as dd00dd
* aqua as 31dddd doesn't look noticeably different to me
* fuchsia as dd31dd doesn't look noticeably different to me.

I think they're probably fine as they are, but if you think any of them are overly bright then you could try reducing them as above.

NQNStudios:
I don’t think I’ll noodle with the values any more.

PR #537, opened 2025-01-19
Keys can shift the screen

### LINKED ISSUE #482: Arrow key in spell target mode casts the spell adjacent to you
So if you're casting Fireball.... you've just pressed one button (probably thinking it would move the camera) and possibly killed your caster.

NQNStudios:
This makes it so arrow keys/numpad can control screen shift when it's available.

When this preference is enabled, #482 will not be an issue (Fix #482). But the player will also miss out on cases where adjacent targeting would be a good thing. I haven't done anything smart to determine whether one mode or the other is better on a case-by-case basis.

### DIFF
--- a/rsrc/dialogs/preferences.xml
+++ b/rsrc/dialogs/preferences.xml
@@ -43,6 +43,7 @@
 	</group>
 	<text name='misc-head' size='large' relative='pos-in pos' anchor='spd-head' top='30' left='0' width='182' height='17'>Miscellaneous:</text>
 	<led name='nosound' relative='pos-in pos' anchor='misc-head' top='6' left='15'>No Sounds</led>
+	<led name='keyshift' relative='pos-in pos' rel-anchor='prev' top='10' left='0'>Directional keys can shift the screen</led>
 	<led name='repeatdesc' relative='pos-in pos' rel-anchor='prev' top='10' left='0'>Show room descriptions more than once</led>
 	<led name='easier' relative='pos-in pos' rel-anchor='prev' top='10' left='0'>Make game easier (monsters much weaker)</led>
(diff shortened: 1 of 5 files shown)


rsrc/dialogs/preferences.xml
```diff
@@ -43,6 +43,7 @@
 	</group>
 	<text name='misc-head' size='large' relative='pos-in pos' anchor='spd-head' top='30' left='0' width='182' height='17'>Miscellaneous:</text>
 	<led name='nosound' relative='pos-in pos' anchor='misc-head' top='6' left='15'>No Sounds</led>
+	<led name='keyshift' relative='pos-in pos' rel-anchor='prev' top='10' left='0'>Directional keys can shift the screen</led>
```
CelticMinstrel, line 46: I think this should make it more clear that it's swapping one capability with another. Something like "Directional keys shift… instead of…".

CelticMinstrel: Better, but I think a different word than "squares" should be used. Maybe "tiles" or "spaces"?

NQNStudios: done and done

NQNStudios: @CelticMinstrel What if holding shift performed the opposite function so both are always available?

NQNStudios: The dialog could have a parenthetical text explaining this

CelticMinstrel: Sure, why not. The only question would be which modifier to pick… what precedent do we have so far? There's also the option of Control/Command (that is, Command on Mac, Control on non-Mac).

Given how "shift" typically means "extend", it seems like that doesn't really fit this kind of use. I guess "alt" fits better. Not sure whether "control" fits.

NQNStudios: replied to this out-of-thread

PR #543, opened 2025-01-21
Scenedit Quality of life: be helpful when scrolling past boundaries

NQNStudios:
I've implemented 2 features here:
* When scrolling past the boundaries of the current outdoor section, you will get a yes/no prompt which can load the adjacent section for you at the corresponding center position.
* When scrolling past the boundaries (literal, not the changeable rectangle) of the current town, the editor will ask if you want to jump to the town's entrance in the outdoors. If there are more than one, you can choose.

I did this because I need to be able to find town entrances in the built-in scenarios so I can debug things.

### DIFF
--- a/src/game/boe.main.cpp
+++ b/src/game/boe.main.cpp
@@ -440,33 +440,28 @@ static void handle_scenario_args() {
 		}
 		// Try to put the party in an outdoor section from which you can enter the town --
 		// so when you leave, you'll hopefully be in the right place.
-		bool found_entrance = false;
-		for(int x = 0; x < univ.scenario.outdoors.width(); ++x){
-			for(int y = 0; y < univ.scenario.outdoors.height(); ++y){
-				for(spec_loc_t& entrance : univ.scenario.outdoors[x][y]->city_locs){
-					if(entrance.spec == *scen_arg_town){
-						// Very janky but I don't know how else to make it properly load the right sections and set i_w_c
-						while(univ.party.outdoor_corner.x > x){
-							shift_universe_left();
-						}
-						while(univ.party.outdoor_corner.x < x){
-							shift_universe_right();
-						}
-						while(univ.party.outdoor_corner.y > y){
-							shift_universe_up();
-						}
-						while(univ.party.outdoor_corner.y < y){
-							shift_universe_down();
-						}
-						outd_move_party(local_to_global(entrance), true);
-
-						found_entrance = true;
-						break;
-					}
-				}
-				if(found_entrance) break;
+		auto town_entrances = univ.scenario.find_town_entrances(*scen_arg_town);
+		if(!town_entrances.empty()){
+			// TODO a dialog could be shown to choose between multiple entrances,
+			// but maybe that would be janky, because this happens when you're trying to launch
+			// from INSIDE a town.
+			town_entrance_t first_entrance_found = town_entrances[0];
+			int x = first_entrance_found.out_sec.x;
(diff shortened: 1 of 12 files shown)


src/game/boe.main.cpp
```diff
@@ -440,33 +440,28 @@ static void handle_scenario_args() {
...
-						found_entrance = true;
-						break;
-					}
-				}
-				if(found_entrance) break;
+		auto town_entrances = univ.scenario.find_town_entrances(*scen_arg_town);
+		if(!town_entrances.empty()){
+			// TODO a dialog could be shown to choose between multiple entrances,
+			// but maybe that would be janky, because this happens when you're trying to launch
+			// from INSIDE a town.
```
CelticMinstrel, line 447: No need. If there are multiple entrances, the town should already be set up to properly handle them – if they are in different outdoor sections, there would be a "change outdoor location" special node near each viable exit, for example. And if they're using the "set town exits in town settings" feature, that should also correctly reposition them when they leave town.

So, this code only really matters if there is just one town entrance… though it certainly doesn't hurt to take the first if there are multiple, just in case.

src/scenedit/scen.actions.cpp
```diff
@@ -1781,13 +1783,119 @@ void handle_keystroke(sf::Event event) {
...
+
+			if(shift_prompt.show() == "yes"){
+				set_current_out(only_entrance.out_sec);
+				start_out_edit();
+				cen_x = only_entrance.loc.x;
+				cen_y = only_entrance.loc.y;
+				redraw_screen();
+				return;
+			}
+		}else if(town_entrances.size() > 1){
```
CelticMinstrel, line 1863: Couldn't this also take into account the exit location settings in the town settings?

NQNStudios: Oh, yes it could. And those could be added to the string choice dialog

NQNStudios: Although it couldn't guarantee that it put you at the exit coordinate in the right outdoor section, if a multi-entrance town changes the section before you leave via special node.

NQNStudios: @CelticMinstrel Given those considerations I don't think the Town Exits feature would be very reliable/good if I did implement it.

NQNStudios: I guess in the most basic case of only having entrances in one section, it would be nice....

CelticMinstrel: I think the Town Exits feature should be thought of as a "shortcut" for when all the exits are in the same section. You shouldn't use it when there are exits in multiple sections.

Furthermore, in the most typical use-cases, there will be a town at the location of the Town Exit point, so generally those exits will be found by your normal search.

But that said, it's not _necessary_ for a town to exist at the Town Exit point. The second fort in Exile II is a good example of a town that would probably (if implemented in Blades of Exile) use the Town Exits feature to point to a location in the outdoors that's not itself a town.

So, I think a reasonable conclusion can be: if all the town locations you found are in the same outdoor section, check the Town Exit location for the direction you're exiting in and add it if it isn't already in the list.

----

By the way, what happens here if there are _no_ entrances to the town?

NQNStudios:
That makes sense for the exits.

When there are no entrances, nothing happens. I didn’t add a message for it
or anything.

CelticMinstrel:
It just stops scrolling?

NQNStudios:
Yes.

PR #545, opened 2025-01-21
Shop change and talk mode fix

### LINKED ISSUE #216: Sweetgrove: Hammer the Healer selling Invisible Herbs
hamer_the_healer

There are invisible items which can be bought and cost gold, but do not appear in any character's inventory.

### LINKED ISSUE #544: Clicking on special talk buttons also clicks toolbar
Image

When I click "Sell", the buffer prints "Look: Select a space".
When I click "Buy", there are 2 click sounds.
When I click "Record", there are 2 click sounds.

I believe the click events are getting passed through to the toolbar buttons somehow, although the code looks like they shouldn't be.

This might be a stretch, but I can imagine this being the explanation for other bugs like #142.

NQNStudios:
I went to make a message print when there are no items available for sale (as suggested in #216 and will fix #216). This wasn't very hard to do.

When a special node starts a shop, it will still display the shop if there are no items. We could change that, too, by printing to the buffer? I don't know if we need to, though.

While implementing this I noticed bug #544 and fixed it (fix #544)

### DIFF
--- a/src/game/boe.dlgutil.cpp
+++ b/src/game/boe.dlgutil.cpp
@@ -219,18 +219,17 @@ void end_shop_mode() {
 	}
 }
 
-void handle_shop_event(location p, cFramerateLimiter& fps_limiter) {
+bool handle_shop_event(location p, cFramerateLimiter& fps_limiter) {
 	if(p.in(talk_help_rect)) {
-		if(!help_btn->handleClick(p, fps_limiter))
-			return;
-		give_help_and_record(226,27);
-		return;
+		if(help_btn->handleClick(p, fps_limiter))
+			give_help_and_record(226,27);
+		return true;
 	}
 	
 	if(p.in(shop_done_rect)) {
 		if(done_btn->handleClick(p, fps_limiter))
 			end_shop_mode();
-		return;
+		return true;
 	}
 
 	for(short i = 0; i < 8; i++) {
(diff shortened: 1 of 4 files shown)


src/game/boe.dlgutil.cpp
```diff
@@ -881,8 +888,10 @@ void handle_talk_node(int which_talk_entry) {
 			return;
 			
 		case eTalkNode::SHOP:
-			start_shop_mode(b,a,save_talk_str1);
-			can_save_talk = false;
+			if(!start_shop_mode(b,a,save_talk_str1,true))
+				save_talk_str1 = "There is nothing to buy from " + save_talk_str1 + ".";
```
CelticMinstrel, line 892: It looks like shops don't use the second text string from the dialogue node. We could use that here, maybe? Although it may be better to use it as a string shown when shopping concludes… I'm not sure.

(Whether we use it here or not, providing this is a reasonable default.)

NQNStudios: I think it'd be nicer to customize what a shopkeeper says when they can't offer you anything, than when you finish shopping. When there's nothing to buy they could apologize, say goods are in a shortage, other interesting and specific stuff. Healers could say "You look fine to me!" and stuff like that.

NQNStudios: I've made the change.

PR #548, opened 2025-01-24
Better editing for talk node keys

NQNStudios:
This is a more involved change than anything in my other scenedit PR.

I really don't like the way the 2 talk node keys are handled in the editor. Filling the fields with 4 spaces, expecting the designer to delete them and put in 4 other characters... just awkward. Also, 'purc' is impossible to remember and I never will.

New behavior is:

* The fields start out empty, and it is *allowed* to keep *one of them* empty. (A node only needs one key to be useful, right?
* When the node is saved, empty strings are replaced with `"    "`. If this is not harmless to the game logic, then I'll need to make the game aware that either key might be `"    "`.
* You cannot type more than 4 characters in these fields (unless there's a way to insert characters I'm not aware of).
* If you put 'buy' in a field, it automatically becomes 'purc'.
* There is a note next to the response keys (kind of squished in the corner) enumerating the special logic of the in-game Buy and Sell buttons.
* My other scenedit PR should be merged first, because these notes overlap with a message that I removed in the other PR. If both PRs are merged it will look like this:

Screenshot 2025-01-23 at 8 09 12 PM

### DIFF
--- a/src/dialogxml/widgets/field.hpp
+++ b/src/dialogxml/widgets/field.hpp
@@ -90,6 +90,10 @@ private:
 	rectangle text_rect;
 	std::vector<snippet_t> snippets;
 	int ip_row, ip_col;
+	/// Setting maxChars AFTER the player has a chance to type in the field would be a bad idea.
+	/// So would calling setText() with a string longer than maxChars.
+	/// Really, it should probably only be set in xml, as the attribute "max-chars".
+	int maxChars = -1;
 	friend class aTextInsert;
 	friend class aTextDelete;
 };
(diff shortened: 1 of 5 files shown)


NQNStudios:
And I just realized we might as well de-capitalize letters automatically. So that's done.

NQNStudios:
Copy/Paste is something worth making sure can't violate the new maxChars limit in field.cpp

NQNStudios:
I just tested copy/paste on Mac in the max-chars fields... and discovered that copy/paste isn't working in *any* text field. So I'll make a note of that somewhere else.

src/dialogxml/widgets/field.hpp
```diff
@@ -90,6 +90,10 @@ class cTextField : public cControl {
 	rectangle text_rect;
 	std::vector<snippet_t> snippets;
 	int ip_row, ip_col;
+	/// Setting maxChars AFTER the player has a chance to type in the field would be a bad idea.
+	/// So would calling setText() with a string longer than maxChars.
+	/// Really, it should probably only be set in xml, as the attribute "max-chars".
+	int maxChars = -1;
```
CelticMinstrel, line 96: I wouldn't call it a priority to have a runtime setter for `maxChars`, but I don't think it's something that's difficult to do – it could just truncate the current text to the new maximum length.

Issue #564, opened 2025-02-03
Add a function to quickly preview a dialog layout


CelticMinstrel:
When designing new dialogs, it takes a lot of work getting the positioning exactly right. It would be nice if there was a way to work this out without constantly quitting and restarting the editor.

Though it could be a separate external tool, I think it would make sense to put it in the scenario editor, as using custom dialogs in a scenario is an eventual goal.

We do have an XSD file that transforms a dialog for previewing in a browser, which used to help… but it hasn't been updated in awhile, and if I recall correctly, it was always a little bit off. I think it would be better to have something that uses the same rendering code.

NQNStudios:
I am whipping this up real quick.

NQNStudios:
Finished by bc5f42a?

PR #565, opened 2025-02-03
quick implementation of previewing dialogxml for #564

NQNStudios:


NQNStudios:
@CelticMinstrel If you want something to use for this asap while you're working on dialog stuff, here you go.

I only have it linked up to the game as a debug action, Shift+D Shift-J. You probably want it linked into the scenario editor as an action too? I can do that later but here's the core of the feature.

PR #569, opened 2025-02-04
On Mac, restore original preferences after replay.

### LINKED ISSUE #456: On Mac, preferences from replays seem to be applied persistently
I think I saw this happen, didn't confirm it, and forgot to write it down.

If it is the case, it'll be pretty annoying and worth fixing. We don't want replays to clobber our preferred settings.

It would make sense if Mac is doing it and Windows/Linux aren't, because Mac handles preferences differently.

NQNStudios:
Fix #456

(I tested this and it works.)

### DIFF
--- a/src/game/boe.main.cpp
+++ b/src/game/boe.main.cpp
@@ -1298,10 +1298,19 @@ void Mouse_Pressed(const sf::Event& event, cFramerateLimiter& fps_limiter) {
 }
 
 void close_program() {
-	// TODO: Ultimately we would like to have cleanup happen automatically, negating the need for this function
-	// On the Mac, prefs are synced automatically. However, doing it manually won't hurt.
-	// On other platforms, we need to do it manually.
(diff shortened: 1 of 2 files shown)


CelticMinstrel:
Does it work when the replay had an error and stopped early? Does it work if the replay ran to completion without quitting the game? (Though the latter seems like it might not be possible, unless you manually craft a replay file.)

NQNStudios:
In the latter case (which is definitely possible, if you use Ctrl+C on the command line), it does restore the preferences but only after you close the window manually.

NQNStudios:
In order to make this bullet-proof I think the way to go would be to make the windows/linux implementation of the preferences available to the Mac build when running in replay mode. Because otherwise, the Mac code isn't equipped to store the multi-type preference things in a std::map.

NQNStudios:
I think I'll do it the other way.

PR #571, opened 2025-02-05
Gorgeous text in scaled mode

### LINKED ISSUE #316: Unblur text when UI scaling is enabled
Essentially what this entails is that text is drawn _after_ the scaling has been applied, but with its font size and position scaled by the same amount.

I'm not sure how much work this is. It might be a lot.

NQNStudios:
This will fix #316 when complete.

There are sure to be edge cases where I've missed calling `clear_scale_aware_text()` on a RenderTexture.

Look how good this looks when text is scaled properly:

Screenshot 2025-02-04 at 9 51 07 PM


NQNStudios:
https://github.com/calref/cboe/pull/571/files#diff-8a4bc018ca25a4b695710596ec3b55e65c184f13b22a58b2313670857388db94R138-R139

I think I thought of a better way: subclassing RenderTexture.

NQNStudios:
One more thing I want to do: Make the Done button and other buttons with baked-in text use nice text. Right now they're just mismatched.

CelticMinstrel:
There are only two buttons with baked-in text: the Done button, and the Race/Traits button.

Note that both of them use navy text instead of black. I do wonder if it's worth doing that on _all_ buttons. I'm not sure what legacy BoE did in this area.

The main issue you'd have with eliminating baked-in text on the buttons is the fact that, currently, the `<button>` tag has no line-breaking control, and there's also no word-wrap applied on buttons. One or both of these would be required to avoid baking the text into the Race/Traits button.

NQNStudios:
image
 
Both of the baked-in buttons are actually here on this one screen.

NQNStudios:
3 lines preview:
image

Issue #495, opened 2024-11-26
Extremely out-of-date scenario distribution info


NQNStudios:
https://github.com/calref/cboe/blob/df0c46d616904402d1284400db755e0f646a8e18/rsrc/dialogs/help-distributing.xml#L5

As we start to look toward scenario editor work. I noticed this dialog has some extremely out of date info--it says to mail scenarios to Spiderweb Software by disk.

Now, fixing this actually requires discussion, because we currently don't have a *best* way to distribute a scenario, AFAIK. I want to make that process smooth, and to feed into something resembling #22

Issue #300, opened 2023-01-07
Update all existing dialogs to use the relative positioning system


CelticMinstrel:
There might be some occasional cases where we don't want to, but they should be uncommon.

NQNStudios:
IMO this isn't actually an issue until we actually want to rearrange a dialog. What if we add a note to the top of every un-converted xml file like

`<!-- NOTE: This file should be updated to use the relative positioning system the next time it changes. -->`

and then close this issue?

CelticMinstrel:
It's not a bad idea.

NQNStudios:
My commit [862ac5e](https://github.com/calref/cboe/pull/575/commits/862ac5e2249b3be24d866512bdaf3580e8ef7dd9) notes all the work that would need to be done for this. None of it is urgent, just nice if we need to change things. So we don't need an open issue.

PR #575, opened 2025-02-06
Dialogxml fixes/notes/updates

NQNStudios:
We have 4 help windows in the scenario editor that are very outdated (including #495) -- I added a cautionary dialog that warns the user the info is outdated and has the URL for openboe.com. (This is temporary and these will eventually probably call launchDocs instead of opening a help dialog at all.)

I fixed up the Hints dialog in the main game so it's up-to-date.

I removed an out-of-date Exile 3 reference from help-inventory and a TODO comment that's not relevant anymore because we do have a JOBS button.

Went through and exhaustively added a note about which dialogxmls should eventually get relative positioning if we ever need to change them. This marks the work that would be required for #300, which is not necessarily worth actually doing for the many many dialogs that just might stay the same forever.

### DIFF
--- a/rsrc/dialogs/1str-lg.xml
+++ b/rsrc/dialogs/1str-lg.xml
@@ -1,4 +1,5 @@
 <?xml version='1.0' encoding='UTF-8' standalone='no'?>
+<!-- NOTE: This file should be updated to use relative positioning the next time it changes. -->
 <?xml-stylesheet href="dialog.xsl" type="text/xsl"?>
 <dialog defbtn='done'>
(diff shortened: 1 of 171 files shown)


CelticMinstrel:
Saying it here so I don't forget: I have a few changes to specific dialogs pending, which I'd like to get in before merging this.

CelticMinstrel:
Okay, if you rebase this on the current master, and make sure the Advanced Town Details dialog doesn't include the "change to relative positioning" note, I'll merge this.

NQNStudios:
@CelticMinstrel done.

PR #576, opened 2025-02-06
Mac replay system don't use UserDefaults for replays

### LINKED ISSUE #456: On Mac, preferences from replays seem to be applied persistently

NQNStudios:
Fix #456 

This is a more proper fix for the problem than #569 was.

Mac builds now include all of the text and map-based preference parsing code that windows/linux do. When the game is run as a replay, this code gets used and NSUserDefaults never come into play at all.

I've used some devilish compiler macros to make the code relatively DRY for the universal preferences code to redirect to mac-specific implementations if it's not a replay.


CelticMinstrel:
Are you able to update the Xcode 4 project to or does that require actually using an older computer? (I can do it on my old computer if there's issues.)

NQNStudios:
I can't remember if I've tried doing that before. Just stepped  away from the computer though so I can't try till later.

CelticMinstrel:
The Xcode 4 project doesn't build currently (in Xcode 4 at least) so it's not a priority, but if it's easy then you might as well do it?

NQNStudios:
I'll try it later, and I think there's another variant of the do_load crash that I can easily fix.

CelticMinstrel:
I think this is roughly ready to merge, but I'm probably going to do it manually and squash up some of the commits.

CelticMinstrel:
Merged manually in c1743afc9546f077ab7f18da4c3266904b509810

Issue #110, opened 2018-02-15
When the party is invisible, it seems to make no difference.


daerogami:
It should make the party immune to archery and melee, but not spells.

---
This is an item logged from the To-Do and was originally reported by (@CelticMinstrel?). If you are familiar with or can recreate this bug, please provide details here.

CelticMinstrel:
Just to clarify for anyone less familiar with the game, "invisible" means the Sanctuary spell.

daerogami:
Is this based on the DnD/Pathfinder Sanctuary spell where a monster has to roll for a chance to even be allowed to roll an attack? When I played as a kid and phones hung on walls, I never understood this spell so I never really used it.

CelticMinstrel:
Pretty sure it was the Sanctuary spell, not Dust of Hiding.

NQNStudios:
I thought maybe they were both invisibility.

CelticMinstrel:
They do both use the same status effect, but the possible bug with Dust of Hiding in #153 was that the status effect is not even applied, while this bug is that the status effect does not work as expected.

NQNStudios:
Ok, it looks like sanctuary is correctly blocking melee attacks but not ranged.

<img width="228" alt="Image" src="https://github.com/user-attachments/assets/8bca3cca-d959-4aa7-bb9a-97328aa20202" />

NQNStudios:
This code handles the case for melee attacks:

https://github.com/NQNStudios/cboe/blob/f22f248f4e05c1f3f0c9ef74126b870e1d1b74fb/src/game/boe.combat.cpp#L2805-L2811

This code should handle it for ranged attacks:

https://github.com/NQNStudios/cboe/blob/f22f248f4e05c1f3f0c9ef74126b870e1d1b74fb/src/game/boe.combat.cpp#L3080-L3085

CelticMinstrel:
So does that mean it's working just fine?

NQNStudios:
I don't know how JV intended for the spell to work.

The issue description says:

> It should make the party immune to archery and melee, but not spells.

And actually, neither of these things are true currently. It works *more* for melee than for archery, but both attack types can still land, and are more likely to the higher the monster's level.

Melee attacks are a D100 vs. hit_chance[level / 2].
Ranged attacks are D100 vs. hit_chance[level].

So ranged attacks hit Invisible characters more often. We could divide the level by 2 for ranged attacks and have them behave more similarly.

NQNStudios:
[...]

PR #583, opened 2025-02-09
make sanctuary effectiveness the same vs missile attacks as melee

NQNStudios:
I don't know if this is the right thing to do, but as noted in #110, invisibility is less powerful protecting against ranged attacks than melee. And that doesn't really make sense. If we want to level the playing field, this PR should do that.


NQNStudios:
Trying again with a better fix

Issue #245, opened 2020-01-28
Armor values different from Exile 3


x-qq:
Exile 3 shirt: defend 1 

![e3](https://user-images.githubusercontent.com/55463047/73231780-e9611480-4189-11ea-8da1-d25a23005d23.png)

Openboe shirt: defend **0**

![nope](https://user-images.githubusercontent.com/55463047/73231807-fda51180-4189-11ea-8345-53f2042f6f8c.png)

x-qq:
Orignal windows BoE: defend 0

![origboe](https://user-images.githubusercontent.com/55463047/73698920-18780880-46eb-11ea-8247-a95228f9ce9f.png)

Looks like there was a nerf of shirts between E3 and BoE.

CelticMinstrel:
This is one of the areas where I think it's fine to depart from the original BoE and treat it as fixing a bug from the original game.

If only it were easy to come up with an actual list of these differences…

josefwk:
There are several different issues at play here.

1. In the item info screen, Exile III shows different item stats than Blades of Exile and OpenBoE. Defend 1 versus Defend 0, as shown at the beginning of this bug report.

2. On the item pickup screen, Exile III and OpenBoE show different item stats than Blades of Exile. Blocks 1-1 damage versus Blocks 0-0 damage. In this case, OpenBoE is actually more correct than the original game in how it displays item stats on pickup.

Exile III:
![Image](https://github.com/user-attachments/assets/eb7574da-7375-490c-9945-4c5a15753205)

OpenBoE:
![Image](https://github.com/user-attachments/assets/e32ed59f-ff9a-4b1a-ba03-6d00ea2d6621)

Blades of Exile:
![Image](https://github.com/user-attachments/assets/4a5a6a2a-ca42-4473-9cad-c89a67946d89)

3. In Blades of Exile, Shirts still block 1 damage despite showing Defend 0 and Blocks 0-0 damage. Just a note that the original game had a discrepancy between reported defense versus calculated defense. I tested this with a 5d1 physical damage special where, with all else being equal, wearing a Shirt protected the wearer from 1 damage (expected 0).

Before taking damage:
![Image](https://github.com/user-attachments/assets/ec376564-25db-4a82-90b7-ae2315a43047)

After taking damage:
![Image](https://github.com/user-attachments/assets/cc608a7f-c59a-4add-b5c5-ca91b8049efc)

[...]

PR #584, opened 2025-02-09
Feature flags in the game and scenarios

### LINKED ISSUE #555: Feature flags

NQNStudios:
I've implemented my idea for the feature flags system. The game has feature flags which are recorded and replayed so we can keep legacy behavior the same in replays. As an example, I've fixed the shirt bug mentioned in #245 in a way that should keep old replay behavior the same (not that we have any replays that had the shirt bug in them...)

I also gave scenarios feature flags which we'll eventually be able to use to detect in the editor whether anything needs to be migrated to a new format to leverage a new feature.

Fix #555


josefwk:
Recommend editing `item.cpp` to display the correct defense value as well. Damage calculation is correct but the item pickup window still shows that a Shirt will Block 1-1 damage.

src/game/boe.global.hpp
```diff
@@ -31,4 +31,14 @@ extern std::map<std::string, int> startup_button_indices;
...
 extern std::map<int, std::string> startup_button_names_v1;
 
+extern std::map<std::string, std::string> feature_flags;
+
+inline bool has_feature_flag(std::string flag) {
+	return feature_flags.find(flag) != feature_flags.end();
+}
+inline std::string get_feature_flag(std::string flag) {
+	if(!has_feature_flag(flag)) return "";
+	return feature_flags[flag];
```
CelticMinstrel, line 41:
```suggestion
	std::map<std::string, std::string>::const_iterator iter = feature_flags.find(flag);
    if(iter == feature_flags.end()) return "";
	return *iter;
```

And the same in scenario.cpp.

CelticMinstrel:
It seems like something is missing here. I don't see any code that initializes the feature flags when starting a new scenario.

NQNStudios:
The feature flags are added to the scenario by the editor when it creates it.

https://github.com/calref/cboe/pull/584/files#diff-e01fb704d3efd4e5ba7dd3cad2b5599bde269e0bde79ab25e9296d8c52ff34a6R2957

CelticMinstrel:
But the game only uses the global feature flags. The scenario feature flags have no effect. Plus, setting the feature flags to empty in a new scenario doesn't seem correct.

Or to put it another way, it looks like there is nothing to make the feature flags actually do the thing they're supposed to do, namely making the shirt block 1 damage if you play a legacy scenario and 0 damage if you play a newly-created scenario.

NQNStudios:
Oh, I wasn't concerned with keeping shirt defense at 1 for legacy scenarios. That just seems like a bug.

The feature flag is there for the hypothetical scenario where a replay might exist where the bug is pivotal. (In this case, I don't think one does.)

NQNStudios:
I care more about the feature flag infrastructure than the shirt defense thing.

CelticMinstrel:
> Oh, I wasn't concerned with keeping shirt defense at 1 for legacy scenarios. That just seems like a bug.

It does seem like a bug, but the shirt defense thing also doesn't seem like a good example of when you'd want a feature flag. Isn't the feature flag system supposed to be a generalization of the scenario "legacy" flag? With this implementation it definitely doesn't seem to be doing its job. Maybe we have different ideas of what it's supposed to do though. But, even considering that, it doesn't make any sense to add scenario feature flags that do nothing at all.

NQNStudios:
The scenario editor and the game will both be able to check for scenario feature flags when we do need to add any.

NQNStudios:
I'm going to separate out the armor fix as its own PR.

NQNStudios:
Retooling this for a different PR.

PR #587, opened 2025-02-09
Fix some ranged attack animation bugs

### LINKED ISSUE #97: Graphical bug when missile trajectory exceeds current view.
When a missile flies off the edge of the terrain view (like if an enemy archer is shooting at your PC and your PC is not within the 9x9), the missile continues in that direction, leaving artifact missiles in its path.

---
This is an item logged from the To-Do and was originally reported by ADoS. If you are familiar with or can recreate this bug, please provide details here.

### LINKED ISSUE #582: spell blast was drawn off-screen
In combat I saw a blast for Ice Bolt on one of my PCs get drawn out-of-bounds of the terrain window. For one thing, it seems like the camera should show the target of the spell? And for another, blasts shouldn't draw outside the border of the terrain window.

NQNStudios:
I need to keep testing this.

Fix #582 
Fix #97

### DIFF
--- a/src/game/boe.newgraph.cpp
+++ b/src/game/boe.newgraph.cpp
@@ -361,10 +361,37 @@ void do_missile_anim(short num_steps,location missile_origin,short sound_num) {
 		return;
 	}
 	
-	if(std::all_of(store_missiles, store_missiles + 30, [](const store_missile_type& m) {
-		return m.missile_type == 0;
-	})) return;
-	
+	// Eliminate missiles traveling 0 distance
+	for(short i = 0; i < 30; i++) {
+		if((store_missiles[i].missile_type >= 0) && (missile_origin == store_missiles[i].dest))
+			store_missiles[i].missile_type = -1;
+	}
+
+	std::vector<location> missile_targets;
+	std::vector<int> missile_indices;
+	int tracking_missile = -1;
+	location camera_dest;
+	for(short i = 0; i < 30; i++)
+		if(store_missiles[i].missile_type >= 0){
+			missile_indices.push_back(i);
+			missile_targets.push_back(store_missiles[i].dest);
+		}
+
+	if(missile_targets.empty()) return;
+
+	if(missile_targets.size() == 1){
+		tracking_missile = missile_indices[0];
+		camera_dest = between_anchor_points(missile_targets[0], missile_origin);
+	}else{
+		std::vector<location> dest_candidates = points_containing_most(missile_targets);
+		camera_dest = closest_point(dest_candidates, missile_origin);
+		tracking_missile = missile_indices[closest_point_idx(missile_targets, camera_dest)];
+	}
+
+	// Start the camera as close as possible to containing the origin and the camera destination
+	// on the same screen
+	center = between_anchor_points(missile_origin, camera_dest);
+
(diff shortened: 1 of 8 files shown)


src/game/boe.newgraph.cpp
```diff
@@ -120,6 +120,15 @@ rectangle explode_place_rect[30];
 
 char last_light_mask[13][13];
 
+static void terrain_screen_rects(rectangle& old_bounds, rectangle& new_bounds) {
+	rectangle rect = rectangle(terrain_screen_gworld);
+	rect.bottom -= 10; // Adjust for pointing buttons
```
CelticMinstrel, line 125: Wait… weren't the point buttons removed from the terrain screen sheet? We probably don't even need this line anymore?

NQNStudios:
Recentering on the target after the missile goes off the edge seems to work... but there are some cases where recentering *before* the animation starts, on a tile between the origin and the destination, could make it so the whole animation fits on the screen. I think that should be the first choice behavior.

src/game/boe.newgraph.cpp
NQNStudios: I got that it was the same thing, just not that it was the blue arrows.

CelticMinstrel: Those arrows used to be attached to the bottom of terscreen.png. They aren't anymore.

NQNStudios:
I'm ready to recommend merging this. There is one more bug I want to fix in ranged animations but it already existed without my changes, eludes fixing after many attempts to debug, and I don't want to let the fixes in this branch go stale because I can't figure that one out at the moment.

This video shows the current master branch running combat with a variety of ranged animations at Slow game speed:

https://youtu.be/I6P-Gy1JARk

It demonstrates #97, #582, and #602 (the bug I mentioned I haven't fixed yet.)

This video shows my updated animation code running the replay of the same combat sequence:

You can see #97 and #582 fixed.

There's also behavioral changes in how the view is framed:

|old|new|
|---|---|
|When player missile anims start, the missile flies from off-screen if the current camera view doesn't show the active PC|When a player missile animation starts, the camera always moves to show the player's attack/cast animation and the origin of the missile|
|When enemy missiles fly off-screen, the camera stays in the same place so you often don't see where the missile lands.|For both player and enemy missile attacks, the camera will pan to a location which better showcases the missile's destination before it lands and the boom happens. This pan happens halfway through the missile's flight, OR when the missile being tracked goes off-screen.|
|When a PC fires multiple missiles, the camera stays where it was when you select the last target. You might not see the majority of your missiles land.|When a PC fires multiple missiles, the game calculates a frame that shows the highest amount of targets and their booms. A missile heading toward the desired frame is tracked, and the pan will happen early if this missile goes out of frame earlier than halfway through the animation duration. (Now, it could be argued that it should always track the last missile that's targeted, but even so, I think it should aim to show as much carnage as it can.|

NQNStudios:
I can't currently stage the replay for you to conveniently run it yourself because of the problem with test scenarios mentioned in #601. However, here is the Ranged Arena scenario compressed:

[ranged-arena.zip](https://github.com/user-attachments/files/18804152/ranged-arena.zip)

PR #598, opened 2025-02-10
Fix restore

### LINKED ISSUE #597: After death, When restoring a save file that's in a scenario, you go to the startup menu
It appears we're still caught in a tangled web with game-loading logic. The change that fixed the scrollbars when restoring (https://github.com/calref/cboe/pull/560) introduced this.

NQNStudios:
I tested that this does not bring back the scrollbar bug. Fix #597

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,7 @@ proj/vs*/*/ipch/
 proj/vs*/.vs
 proj/vs*/.vs
 proj/vs*/packages/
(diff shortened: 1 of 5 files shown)


CelticMinstrel:
I think this debug action should ask for a confirmation. If someone's testing their scenario and accidentally hits X instead of C, they won't be a happy camper.

PR #599, opened 2025-02-10
His/her -> their, He/she -> they

NQNStudios:
I found more pronouns that can be simplified.


CelticMinstrel (commented):
Seems like there are no reflexive cases of this.

PR #601, opened 2025-02-13
.boes file needs to exist to launch replay test scenario

NQNStudios:
One of my test replays won't actually run with a fresh clone, because the packed `.boes` file is needed for the game to launch a scenario. So I've added the packed file for that replay test scenario.


NQNStudios:
Oh, this is more complicated than it seems. Adding the boes file to the repo changes the scenario choice list to have 2 of the same scenario, only one of which works and launches the scenario.

Modifying the game to be able to launch an unpacked scenario might be better because then the scenario editor could "Launch" unpacked versions of the built-in scenarios.

CelticMinstrel:
Hmmmm…

I think it makes sense for the scenario picker to only list packed scenarios, because that's the form a scenario should be installed in. But if you specify a scenario on the command-line, the scenario picker shouldn't be involved at all, so it should be able to launch any scenario regardless of location or whether it's packed.

I'm not sure how this relates to your replay test scenario however.

NQNStudios:
Opening an issue instead of a draft PR.

PR #604, opened 2025-02-14
Fix a help dialog that appears in the top-left

NQNStudios:
This resolves a TODO note explaining a bug I've confirmed and fixed.

### DIFF
--- a/src/game/boe.main.cpp
+++ b/src/game/boe.main.cpp
@@ -962,7 +962,7 @@ void init_boe(int argc, char* argv[]) {
 		plop_fancy_startup(fps_limiter);
 	
 	cUniverse::print_result = iLiving::print_result = add_string_to_buf;
-	cPlayer::give_help = give_help;
+	cPlayer::give_help_enabled = true;
 	init_fileio();
 	init_debug_actions();
 	init_spell_menus();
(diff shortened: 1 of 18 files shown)


NQNStudios:
This doesn't work because give_help doesn't exist for the scenario editor.

give_help is literally copy-pasted between the game and pceditor code, so while I'm at it, I'll just DRY that up.

CelticMinstrel:
Let's also make `help_forced` be a parameter, instead of magically adding 200 to one of the strings. It might be that the original BoE only had 200 strings in each resource, and it might have even been done that way for a very good reason (if the counter was a byte), but the current format has no restrictions on size.

src/game/boe.main.cpp
```diff
@@ -1453,10 +1462,10 @@ void handle_menu_choice(eMenu item_hit) {
 			break;
 		case eMenu::ABOUT_MAGE:
 		case eMenu::ABOUT_PRIEST:
-			menu_give_help(209);
+			menu_give_help(9);
 			break;
 		case eMenu::ABOUT_MONSTERS:
-			menu_give_help(212);
+			menu_give_help(12);
```
CelticMinstrel, line 1468: Looks like you forgot to add `true` here.

PR #615, opened 2025-02-15
Fix line breaks being ignored. Fix #608

### LINKED ISSUE #608: Dialog line breaks not working (sometimes?)
The `<br/>` tag seems to be ignored. I tried to add one to the "Default Attitude" text in the Edit Monster dialog, but the text still renders all on one line.

I'm pretty sure I've seen them work in other places though. There's probably a setting that makes them break, which is wrong. Either they should work in any text, _or_ you should get a helpful error when using them in a text that doesn't support them.

NQNStudios:
Fix #608

I noticed 'Number of Dice' and 'Number of Sides' texts need to be widened in edit-monster--is that something you're working on or should I push my fix?

### DIFF
--- a/rsrc/dialogs/edit-monster.xml
+++ b/rsrc/dialogs/edit-monster.xml
@@ -42,8 +42,8 @@
 	<field name='priest' top='226' left='206' width='51' height='16'/>
 	
 	<text size='large' top='82' left='273' width='158' height='16'>Hand to hand combat:</text>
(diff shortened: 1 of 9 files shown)


CelticMinstrel:
I already pushed my edits to edit-monster, so go right ahead.

NQNStudios:
Fixing the line breaks bug is gonna require checking all the dialogs again...

I'm gonna make a debug action that just previews EVERY dialog one by one. Because with other changes in the mix like #571 there's no guarantee that it won't be useful for future tedious reformat-checking.

PR #616, opened 2025-02-15
DRY advanced town help. Fix #611

### LINKED ISSUE #611: DRY advanced town details help
The code to show help strings in the Advanced Town Details is more or less the same as the `give_help` function. The only difference is the file it loads the strings from. Since these are scenario editor help strings, I thought it wouldn't be a good idea to put them in the main help strings file.

NQNStudios:
Adding a global variable when I know we want to have less of them. 😬

### DIFF
--- a/src/scenedit/scen.townout.cpp
+++ b/src/scenedit/scen.townout.cpp
@@ -40,6 +40,7 @@ extern cScenario scenario;
 extern cOutdoors* current_terrain;
 extern location cur_out;
 extern cUndoList undo_list;
+extern std::string help_text_rsrc;
 
 const char *day_str_1[] = {"Unused","Day creature appears","Day creature disappears",
 	"Unused","Unused","Unused","Unused","Unused","Unused"};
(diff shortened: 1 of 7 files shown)


src/scenedit/scen.townout.cpp
```diff
@@ -808,10 +808,9 @@ void edit_advanced_town() {
 		auto iter = std::find(help_ids.begin(), help_ids.end(), which);
 		if(iter == help_ids.end()) return true;
 		int n = iter - help_ids.begin();
-		std::string str = get_str("town-advanced-help",1 + n);
-		cStrDlog display_help(str,"","Instant Help",24,PIC_DLOG, &me);
-		display_help.setSound(57);
-		display_help.show();
+		extern std::string help_text_rsrc;
+		help_text_rsrc = "town-advanced-help";
+		give_help(1 + n, 0, me, true);
```
CelticMinstrel, line 813: You missed the second case just a few lines down.

PR #617, opened 2025-02-16
Fix pcedit gold limit & remove os string

### LINKED ISSUE #79: Character editor does not let you have more than 25000 gold
Isn't the limit 30000-something?

---
This is an item logged from the To-Do and was originally reported by ADoS. If you are familiar with or can recreate this bug, please provide details here.

### LINKED ISSUE #522: Remove OS string from scenario format
Is there even any reason for this information to be included? I honestly don't remember why I put it in the scenario format in the first place. If it was to distinguish Windows from Mac scenarios, that isn't a good method at all (and anyway is totally unnecessary since the new scenario format is text-based).

NQNStudios:
Remove give/take gold/food function redefinitions from the PC Editor where they weren't being used anyway. DRY up the ones in the game code.

Make the PC editor's gold limit match the game's gold limit. Fix #79 

Unrelatedly, Remove the `<os>` string from scenarios (but don't make validation fail for old ones). Fix #522

### DIFF
--- a/rsrc/bases/bladbase/scenario.xml
+++ b/rsrc/bases/bladbase/scenario.xml
@@ -25,7 +25,6 @@
     <creator>
         <type>oboe</type>
         <version>2.0.0</version>
-        <os>Microsoft Windows 7 Service Pack 1</os>
     </creator>
     <game>
         <num-towns>1</num-towns>
--- a/rsrc/scenarios/busywork/scenario.xml
+++ b/rsrc/scenarios/busywork/scenario.xml
@@ -40,7 +40,6 @@
     <creator>
         <type>oboe</type>
         <version>2.0.0</version>
-        <os>Apple Mac OS X Version 10.7.5 (Build 11G63)</os>
     </creator>
     <game>
         <num-towns>4</num-towns>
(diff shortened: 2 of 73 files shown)


NQNStudios:
I had accidentally left some defines I added as an experiment that weren't actually used. Just got rid of those.

CelticMinstrel:
Why do old scenarios still need to validate?

NQNStudios:
I have test scenarios floating around in boes files.

NQNStudios:
When I try to launch them it throws an error message, which is inconvenient.

CelticMinstrel:
What exactly throws an error message? Is it from xmllint or is it from the game itself?

NQNStudios:
The game does.

CelticMinstrel:
The game doesn't even use the schema, so if the error comes from the game, leaving the tag in the schema won't do anything.

NQNStudios:
Oh. I must have made a wrong assumption (based on an error that gets thrown when a scenario with <feature-flags> in it, is loaded from a non-feature-flags build), that there *would* be an error encountering the <os> tag unless I kept it in there.

PR #620, opened 2025-02-16
Spec strings

NQNStudios:
I really didn't like the get_strs() function in boe.special.cpp which always had to get 2 strings at a time even if you only needed 1 or were going to need 6 total. I made all that code DRY and less horrifying.

I also fixed some crash bugs in various string dialogs.

### DIFF
--- a/src/game/boe.items.cpp
+++ b/src/game/boe.items.cpp
@@ -590,7 +590,7 @@ short custom_choice_dialog(std::array<std::string, 6>& strs,short pic_num,ePicTy
 	
 	std::vector<std::string> vec(strs.begin(), strs.end());
 	// Strip off trailing empty strings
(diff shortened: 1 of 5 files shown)


CelticMinstrel:
This is broken. The scenario string list will be wiped whenever fetching a town or outdoors string.

I also don't like the name `get_scen_str`. It's not fetching a scenario string, necessarily – it depends on `cur_type`.

NQNStudios:
what about get_context_str()?

CelticMinstrel:
Not a fan of that either, but it's better than `get_scen_str`.

I also looked at the two crash fix commits, and they look fine.

NQNStudios:
I put the function names back to `get_str` and `get_strs`.

PR #622, opened 2025-02-17
Write timer index. Fix #621

### LINKED ISSUE #621: Read/write code for timers doesn't match
On write it outputs 3 numbers but on read it expects 4 numbers.

NQNStudios:
It appears this fix is incredibly simple. Am I missing anything?

Fix #621

### DIFF
--- a/src/universe/party.cpp
+++ b/src/universe/party.cpp
@@ -809,7 +809,7 @@ void cParty::writeTo(cTagFile& file) const {
 	for(unsigned int i = 0; i < party_event_timers.size(); i++) {
 		if(!party_event_timers[i].is_valid()) continue;
 		auto& timer_page = file.add();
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
> It appears this fix is incredibly simple.

Pretty much. The only complexity is the question of whether the "read" form or the "write" form is the better way to do things.

NQNStudios:
The write form is eliminating empty spaces in the timer vector and dropping their indices. That seems like a bad thing -- isn't the index of the timer the identifier by which its events know to track it?

NQNStudios:
Oh, actually it looks like each entry contains the special node that the timer calls. So maybe the other way would be fine, too.

CelticMinstrel:
Hmm. I'm not aware of anything that _currently_ relies on the index (from what I recall, timers are evaluated by simply iterating over the list). But I could imagine we may want things in the future that rely on the index, so preserving it doesn't seem bad.

PR #632, opened 2025-02-18
Redo scen details

### LINKED ISSUE #593: Redo & split up Scenario Details
It should be in a stack so I can see/edit the whole thing with my preferred UI scale.

NQNStudios:
REWORKED: A feature flag denotes whether a scenario uses the new or old format for its metadata. With the old format (legacy scenarios and openBoE scenarios created prior to this PR) everything displays in the scenario picker as before, and scenario designers get 2 text fields to display however they want, and 2 text fields that are pretty much useless.

With the new format, Author and Contact info are formatted onto the first line in the scenario picker display. The scenario designer gets 1 line to write a teaser.

Fix #593

Also fix a bug where scenario ratings were appearing as integers ingame instead of the correct "G", "PG", etc.


CelticMinstrel:
I guess I don't really get it. You're taking the author and contact info and stuffing them into "who wrote 1" in a specific format. If you want them in a specific format, why stuff them into a freeform field? I guess it does let the author decide to use a different format if they want, but what's the benefit of that?

I don't remember where the Who Wrote strings were displayed in legacy BoE (if at all), so it's hard to say if this change to their "identity" makes sense. Aren't they shown in the scenario select dialog?

NQNStudios:
They're shown in the scenario selection dialog, yes. And the reason I prefer suggesting the format they should take, as opposed to actually forcing it, is that forcing it would require messing with the dialogxml for the scenario picker, and changing the scenario picker's code, too. I'd rather not take the risk of introducing new bugs doing that for little gain.

NQNStudios:
Oh, and it keeps legacy scenarios displaying Who Wrote1 and Who Wrote 2 the way they intended.

NQNStudios:
It is a little wonky. I'll rework it.

NQNStudios:
@CelticMinstrel reworked, with the PR description updated to explain the changes.

NQNStudios:
I forgot to cherry-pick the menubar item commits

NQNStudios:
There we go.

rsrc/dialogs/edit-scenario-details.xml
```diff
@@ -2,30 +2,39 @@
...
 		This is where you can define the various pieces of information the user will see when deciding whether or not to play your scenario.
 		The meanings of all these fields are given in the documentation.
 	</text>
-	<text size='large' top='6' left='50' width='256' height='17'>Scenario Details</text>
-	<text top='81' left='50' width='120' height='14'>Version number:</text>
-	<text top='102' left='50' width='120' height='14'>Credits, Part 1:</text>
-	<text top='150' left='50' width='120' height='14'>Credits, Part 2:</text>
-	<text top='213' left='50' width='120' height='14'>Contact Information:</text>
+
+	<text name='title-text' top='65' left='50' width='120' height='14'>Title:</text>
```
CelticMinstrel, line 13: I think the ID (`name`) for these various `<text>` elements should end in something other than `-text` – to me, that implies it should hold the actual title, not a label for the title. `-label` is the most obvious alternative. This applies to all the `<text>` elements that gained a new ID.

CelticMinstrel:
Merged manually in 0f4b112 and 0afed5d.

PR #633, opened 2025-02-18
when dumpfound/sleep status is 0 bound low to 0. Fix #138

### LINKED ISSUE #138: There's a potential issue with items that cure sleep or dumbfounding
They can also be used to gain hyperactivity or enlightenment; this is a departure from original behavior.

---

NQNStudios:
There already was a check to prevent this bug. I don't fully understand it.
Changing `>` to `>=` fixes the problem. But does it break these statuses in other contexts?

Fix #138

### DIFF
--- a/src/universe/living.cpp
+++ b/src/universe/living.cpp
@@ -40,7 +40,7 @@ void iLiving::apply_status(eStatus which, int how_much) {
 	if(which == eStatus::ASLEEP || which == eStatus::DUMB) {
 		// No "wrapping" allowed for these effects.
 		if(status[which] < 0) hi = 0;
-		else if(status[which] > 0) lo = 0;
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
> But does it break these statuses in other contexts?

Probably test the Hyperactivity spell? That's the only thing that adds negative sleep as far as I'm aware.

For negative dumbfounding, there's a new item spell.

I don't think items are capable of inflicting hyperactivity or enlightenment to begin with, actually. That might be a flaw in the "Affect status effect" ability… but I suppose you could try creating such an item as a test.

My suspicion is that this could indeed break things like Hyperactivity.

NQNStudios:
I found hyperactivity still works.

PR #634, opened 2025-02-18
Fix another crash on page 2 of spell picker

NQNStudios:
Page 2 of the spell picker has always been touch-and-go... I found a new crash today because of trying to assign an LED status to a button which would be hidden on page 2. We simply can say that a PC cannot cast the None spell, to fix this.


CelticMinstrel:
I think the NONE spell is actually used somewhere – probably the "Start Targeting" special node? We should make sure that this change doesn't break special targeting. (It might be fine if the "can cast" check isn't made in that context.)

NQNStudios:
I don't see the `can_cast()` check being called for special node targeting.

But also, I found that special node targeting isn't working even without this change (#636)

PR #639, opened 2025-02-20
Debug actions

### LINKED ISSUE #635: Debug give item gives to first PC, not active combat PC

NQNStudios:
Debug actions changes, some of which served as feature flags examples. 

Fix #635


NQNStudios:
New commit uses a `#define` simply to remove some mind-numbing repetition from the massive `replay_action()` function.

CelticMinstrel:
I get the intent of that macro, but I don't really like it… I'm pondering maybe making a dispatch-based system similar to the debug actions or the dialogxml pict renderers? Then you'd write functions like this:

```cpp
DEFINE_REPLAYABLE_FUNCTION(handle_pick) {
	location destination = location_from_action(action);
	handle_bash_pick(destination, did_something, need_redraw, false);
}
```

And it would expand to something like this (somewhat pseudo-code):

```cpp
static void handle_pick(const Element& action, bool& did_something, bool& need_redraw, bool& need_reprint) {
	// Avoid "unused" warnings
	(void) did_something; (void) need_redraw; (void) need_reprint;
	location destination = location_from_action(action);
	handle_bash_pick(destination, did_something, need_redraw, false);
}
static_replay_registrar_t __unique_name__("handle_pick", handle_pick);
```

Where `static_replay_registrar_t` (not the best name) is a struct whose constructor adds that function to a global map. Then for the trivial ones you can probably have a second macro that simply forwards, ignoring the XML element reference, to the overloaded function by the same name that doesn't take that parameter.

NQNStudios:
I'm bearing down on bugfixes/top priorities, so this macro is all the work I'm willing to put into cleaning that part up for now. Feel free to take the debug actions and leave the macro if you want.

Issue #607, opened 2025-02-15
Rework special node button dictionary


CelticMinstrel:
The special node `button_dict` (in special.cpp) encodes what sort of button should be placed on each special node field and what kind of picker should appear when that button is pressed. It's encoded as 2-dimensional arrays with each ASCII character having a different meaning. It's quite compact and works just fine, but actually editing it to fix issues or add a new type is rather painstaking.

I'm trying to think of a better way to do it that's easier to edit while not being overly verbose. This is in service of eventually adding several new picker types (including some dynamic pickers), as well as a "simple toggle" case which replaces the entire field with an LED.

Ultimately what it needs to do is to populate a `node_properties_t` struct for each possible special node type. Directly declaring the array of `node_properties_t` would probably be possible but feels like it would be overly verbose.

NQNStudios:
Oh dear... Somehow I'd never seen this before now.

https://github.com/calref/cboe/blob/e83ad326188a3cd13ed6a524efcc4c4013d1be9b/src/scenario/special.cpp#L583

My gut reaction is that I hate it.

NQNStudios:
What if the information was encoded as XML, which was written by a WYSIWYG editor showing what the edit special node dialog would look like when the node's definition was parsed?

CelticMinstrel:
That sounds like extreme overkill for something that's only going to be edited by developers… it's not the most terrible of ideas, but I don't think I like it.

CelticMinstrel:
I got started on trying to clean this up. The above commit defines the structure of how the information should be in memory, while retaining the magic characters during loading.

Any comments are welcome.

NQNStudios:
I was baffled recently by trying to parse through this area of code the other day:

https://github.com/calref/cboe/commit/3bd4adcf6a7b8b4c706b3d8d72592a8144c532eb#diff-176a6d3d8f719743296a0862725f49ddc4706bde78837e720bec1ab78b9c6200L880

And I can say that it does look a lot better with enums than with magic characters as the case labels.

CelticMinstrel:
Does it seem like the above WIP commit might be going in the right direction? (Pay the closest attention to the newly-added file.)

NQNStudios:
Seems like you accidentally deleted the weight field from item.hpp?

But I really like the way it's going, with the builder pattern. It's much more readable to see which node has which properties.

[...]

PR #640, opened 2025-02-20
Allow animated pictures in special dialogs

### LINKED ISSUE #146: Dialog node with boom graphic only lets you select static booms

NQNStudios:
This adds the choice to make pictures in special dialogs animated. (Town Stairs are not included because they don't have a free node flag to match the other ones. But they could get this feature too, and if they did, `anim_pict` could be removed.)

I tried to change it so -2 is the magic value for an infinite anim loop but that I haven't tested that thoroughly.

Fix #146


NQNStudios:
Ah yes. Choosing an animated boom for that node cause the game to crash.

I just tested and found that Do SFX burst isn't working... Gotta check if this PR broke it or if it was already broken.

NQNStudios:
I've added a restriction on Animate Fake Damage so it's static booms only.

This required modifying the special nodes dictionary so it could conflict with your WIP #607 branch.

NQNStudios:
I tested the -2 = infinite change and it works. Now the `-1` default value = no animation.

NQNStudios:
@CelticMinstrel win-vs64 will fail but this should be ready.

PR #643, opened 2025-02-20
Fix text appearing out of alignment with adjusted booms

### LINKED ISSUE #127: Damage burst appears on nose of bear when it is damaged, instead of in the middle.
---
This is an item logged from the To-Do and was originally reported by Tados. If you are familiar with or can recreate this bug, please provide details here.

NQNStudios:
This fixes the rendering of melee damage booms on enemies larger than 1x1. It also doesn't break spell booms (which already worked).

I saw a place to do a refactor which I thought would make it easier to debug this. The fix (in the second commit) could be cherry-picked without the refactor if it looks like I got that wrong.

Fix #127

### DIFF
--- a/src/game/boe.combat.cpp
+++ b/src/game/boe.combat.cpp
@@ -510,6 +510,7 @@ void char_stand_ready() {
 	univ.current_pc().ap = 0;
 }
 
+// Melee attack a monster
 void pc_attack(short who_att,iLiving* target) {
 	short r1,r2;
 	short hit_adj, dam_adj;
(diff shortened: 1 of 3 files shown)


src/game/boe.combat.cpp
```diff
@@ -4030,13 +4021,16 @@ bool monst_cast_priest(cCreature *caster,short targ) {
 	return acted;
 }
 
-void damage_target(short target,short dam,eDamageType type,short sound_type) {
-	if(target == 6) return;
+short damage_target(short target,short dam,eDamageType type,short sound_type, bool do_print, short who_hit, eRace race) {
+	if(target == 6) return 0;
```
CelticMinstrel, line 4025: Come to think of it, I guess this should exclude everything from 6 up to (but not including) 100…

Issue #183, opened 2018-03-22
No mouse/keyboard inputs accepted under native Linux build


seisatsu:
Hey, so I'm not sure about the general state of Linux support, but I got cboe to compile and run under Linux and X11 on the first try. Unfortunately, once starts it won't accept any mouse clicks or keyboard inputs, including window manager keybinds; since it is fullscreen it essentially hangs X11 at the main menu and I have to switch virtual terminals to kill it.

CelticMinstrel:
Thinking about it, it's possible this is actually the "dialog appears behind main window" bug... when I first tried it on Linux, I manually forced it to windowed mode by editing the preferences file.

murlock:
What is your distro please ?

With Xubuntu 17.10, I don't encountered any issue with X11, keyboard or mouse (but I suspect it was false fullscreen as taskbar was still shown)

CelticMinstrel:
It is indeed false fullscreen (a borderless screen-filling window).

CelticMinstrel:
I think the window manager would need to be able to report a false screen size value in order to not cover the taskbar? Not sure.

Linux (and Windows) versions store preferences _~/.oboe/blades prefs.ini_ if I recall correctly. I think the pref you'd need is something like `DisplayMode=6` but I can't quite remember (the value is almost certainly correct, key might be wrong).

The prefs file may also not exist if the game has not yet been run and shut down gracefully. Just create it in this case.

seisatsu:
@CelticMinstrel the prefs file did not exist, and creating it and adding the line did not work. Can you please confirm the name and contents of your prefs file?

CelticMinstrel:
Apologies for the misinformation; I was attempting to look through the source on my phone to remind myself of the correct information, and ended up getting confused in part due to the difficulty of viewing source code on the phone. Even the value, the one thing I was so sure about, was incorrect (but the key is correct.

The file is not a true ini format, just hand-rolled key=value data, so there are no sections. So all you need to do is add the line `DisplayMode=5` to _.oboe/blades/bladesprefs.ini_ and that should enable windowed mode.

(I don't remember why I had that nested _blades_ directory in there, it seems a bit weird now. Maybe something like just in case some other program decided to use _.oboe_ for something? I dunno.)

seisatsu:
Results in a window opening with the Spiderweb Software splash for a fraction of a second followed by:
```
[seisatsu@voidstar Blades of Exile]$ ./Blades\ of\ Exile
[...]

PR #649, opened 2025-02-22
linux: let SFML handle makeFrontWindow()

NQNStudios:
Well, the last PR which I made into a draft turned into a big disappointment. But I did notice one thing, which is that Linux does *nothing* to try to bring the right window to the front. So I just added a call to SFML's function for that. Maybe this fixes issue #183.


NQNStudios:
This one is so trivial but I just can't test it right now, so I'm closing it for the time being. Can reopen when I setup linux again.

PR #652, opened 2025-02-22
Fix scaled text offset in fullscreen modes. 

### LINKED ISSUE #651: font misplaced on title screen
Just tried latest master and text seems to be misplaced on the main screen. tested on Arch Linux with wayland. See attached screenshot for reference. In-game stuff seems to be fine, but barely tested.

Furthermore, the menu stuff at the top is tiny, would be nice to have it scale too, but probably worth another issue.

Image

NQNStudios:
Fix #651

Here's a fullscreen window mode shown fixed:

Screenshot 2025-02-22 at 8 47 43 AM

### DIFF
--- a/src/gfx/render_image.cpp
+++ b/src/gfx/render_image.cpp
@@ -14,6 +14,7 @@
 
 #include "fileio/fileio.hpp"
 #include "gfx/render_shapes.hpp"
+#include "gfx/render_text.hpp"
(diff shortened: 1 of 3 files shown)


NQNStudios:
Oh, no it doesn't. The text in-game just takes a different code path which also needs the fix.

Issue #299, opened 2023-01-07
Add scenario settings to set custom boat and/or horse images


CelticMinstrel:

NQNStudios:
We could also do this for individual boats/horses. A scenario setting could set the default, and each vehicle could have its own. Or if having it in two places is too confusing, it could just be on individual vehicles.

CelticMinstrel:
Generally the solution to this is to make release builds, not debug builds. I wouldn't suggest putting a lot of time into optimizing this.

There's no harm it poking around a bit for obvious stupid things though.

CelticMinstrel:
Okay, but then their issue is not "movement is sluggish in debug builds".

NQNStudios:
Oh, I see how using a monster graphic could be very cool. You could turn any creature into a mount.

CelticMinstrel:
Meanwhile, I was thinking the other way around. You could make a monster that's just a boat.

PR #655, opened 2025-02-23
Update the special node edit dialog with a bunch of new pickers

CelticMinstrel:
The primary goal of this pull request is to add a Choose button (or similar) to as many fields of the Special Node edit dialog as possible. It also reworks the format of the picker dictionary (as per #607) in order to more easily expand its capabilities.


NQNStudios:
I'm really excited about this one for when I get to actually make a scenario.

NQNStudios:
Today I noticed that damage type pickers use boom graphics in the picker, but some booms appear multiple times, I think one didn't have a boom and it didn't really make perfect sense. Maybe worth looking at if you haven't.

CelticMinstrel:
> Today I noticed that damage type pickers use boom graphics in the picker, but some booms appear multiple times, I think one didn't have a boom and it didn't really make perfect sense. Maybe worth looking at if you haven't.

The damage type picker is one of the ones that uses an "augmented" graphic picker. When you select one of the options, a string appears at the bottom left saying what it is. By doing this in the damage picker, you can see that one of the "magic" booms is magic and the other us unblockable. Similarly, the extra two "blood" booms are demon and undead damage.

I can't deny that there could be a better way of doing it, but it functions.

CelticMinstrel:
I think the last big thing I plan to add here is an SDF picker. It would sort of combine aspects of the location picker and the editable string picker.

CelticMinstrel:
I'm marking this ready for review. I do still want to add a few more things:

* [ ] A system to enable special node pickers to depend on the values of other fields.
* [ ] Fix all cases where the special node field prompt overflows the available space, and cases where the prompt is the field name when it should be "Unused".
* [ ] Add help strings for every special node.
* [ ] Either fix the If Statistic? picker so you can select things like Current HP, or add a new node that allows that.
* [ ] Most things in the dialogue node editor don't have proper pickers – fix that.
* [ ] Use the SDF picker in many more places.
* [ ] Remove Edit Saved Item Rectangles. Add a toolbar button instead. (This is instead of making Edit Saved Item Rectangles use the location picker.)
* [ ] Add minor variations of the location picker for rectangles, paths, or other cases where multiple locations are needed.
* [ ] Add a picker for the strings in the Story Dialog special node.

* [ ] Add pickers for pointers. Details still TBD.
* [ ] Something to do with spell patterns TBD.

But they don't necessarily have to be in this PR. Whether they are or not depends on when they get done.

rsrc/dialogs/edit-quest.xml
```diff
@@ -10,20 +10,21 @@
...
 	<text top='307' left='50' width='150' height='16'>Experience Points:</text>
 	<field name='xp' top='306' left='205' width='110' height='16'/>
 	<text top='333' left='50' width='150' height='16'>Gold:</text>
 	<field name='gold' top='332' left='205' width='110' height='16'/>
-	<led name='rel' wrap='true' top='208' left='330'>Deadline is relative to start day</led>
-	<led name='start' wrap='true' top='246' left='330'>Player is given quest when scenario starts</led>
-	<led name='inbank' top='285' left='330'>Include in a job bank:</led>
-	<field name='bank1' top='306' left='344' width='110' height='16'/>
-	<field name='bank2' top='332' left='344' width='110' height='16'/>
+	<led name='rel' wrap='true' top='208' left='345'>Deadline is relative to start day</led>
```
NQNStudios, line 23: This could either mean the scenario start day, or quest start day. I'm not sure which it means.

rsrc/strings/specials-text-general.txt
```diff
@@ -185,10 +196,11 @@ Unused
 0 - allow, 1 - prevent
 Unused
 Unused
-0 - don't force, 1 - force if blocked
+Force allow if blocked?
```
NQNStudios, line 199: Kind of confused on the meaning here.

src/dialogxml/dialogs/strchoice.cpp
```diff
@@ -60,11 +62,17 @@ cDialog* cStringChoice::operator->() {
...
+			strings.resize(per_page * ceil((cur + 1) / double(per_page)));
+		} else cur = 0;
+	} else if(cur < 0) {
+		cur = 0;
+	}
 	page = cur / per_page;
-	fillPage();
 	dlg.setResult<size_t>(selectedIndex);
-	dlg.run();
+	using namespace std::placeholders;
```
NQNStudios, line 74: Is this `using` actually necessary? I don't see a placeholder.

src/dialogxml/widgets/stack.cpp
```diff
@@ -55,25 +55,25 @@ void cStack::draw() {
 	for(auto& p : controls) {
 		p.second->draw();
 	}
-	if(drawFramed) drawFrame(2, frameStyle);
+	drawFrame(2, frameStyle);
```
NQNStudios, line 58: Did you figure out the bug you mentioned with the white lines appearing?

NQNStudios (commented):
The scope of the changes is huge and I think this is taking the scenario editor in a very good direction overall. I still need to look at the actual refactor of the node dictionary.

src/fileio/fileio_scen.cpp
```diff
@@ -904,6 +904,23 @@ void readScenarioFromXml(ticpp::Document&& data, cScenario& scenario) {
...
+					game->GetAttribute("id", &strnum);
+					list.resize(strnum);
+					int i = strnum - 1;
+					Iterator<Element> vehicle;
+					for(vehicle = vehicle.begin(game.Get()); vehicle != vehicle.end(); vehicle++) {
+						vehicle->GetValue(&type);
+						if(type == "name") {
+							vehicle->GetText(&list[i].name);
+						} else if(type == "pic") {
+							vehicle->GetText(&list[i].pic);
```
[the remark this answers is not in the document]
CelticMinstrel: It only adds the storage space for that information in the scenario format – it's not yet used for anything. So yes, it's related to #299.

rsrc/strings/specials-text-general.txt
CelticMinstrel: It's something like "if the other field is set to 'allow', override the terrain's blockage". I _think_ it only applies when the action being prevented is movement, but not quite sure.

Better suggestions for the wording are welcome.

src/dialogxml/widgets/control.hpp
```diff
@@ -348,20 +352,33 @@ class cControl {
...
+	/// @param p The parent dialog, container control, or window.
+	cControl(eControlType t, iComponent& parent);
 	virtual ~cControl();
 	/// Draw the control into its parent window.
 	virtual void draw() = 0;
 	cControl& operator=(cControl& other) = delete;
 	cControl(cControl& other) = delete;
+	/// Sets the positioning method of this control.
+	/// Note: this only has an effect when called from another control's parse function.
+	void setPositioning(const std::string& anchor_id, ePosition h, ePosition v) {
```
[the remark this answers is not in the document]
CelticMinstrel: I don't think so? First of all, nothing could be broken unless this is actually called. Currently, it's only called from the new tilemap widget. I don't think the usage in the tilemap widget can break anything. The tilemap widget fiddles with the names of its nested elements, so essentially all it does is correct the anchor of the nested elements to refer to the adjusted name instead of the name from the XML.

src/dialogxml/widgets/stack.cpp
CelticMinstrel: I did not. I strongly suspect that this is the cause but I didn't get around to investigating in detail.

src/doxy/mainpage.md
```diff
@@ -287,14 +289,32 @@ The `<pane>` tag
...
+It can contain any elements except for nested `<stack>`, `<pane>`, or `<tilemap>` elements.
+
+The `<tilemap>` tag accepts the following attributes:
+
+* `framed` - See **Common Attributes** above. Defaults to `false`.
+* `outline` - See **Common Attributes** above.
+* `rows` - The number of rows to generate. Required.
+* `cols` - The number of columns to generate. Required.
+* `cellspacing` - If specified, adds a buffer of this size between each cell.
+* `<mapgroup>` - Defines an LED group that won't be instanced by the tilemap – that is,
```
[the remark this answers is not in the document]
CelticMinstrel: That's a little unhelpful. Any suggestions for how to improve it? Or if you really don't understand what it's saying, what points can I clarify for you?

NQNStudios:
When an LED is placed next to ex2a, if the value in the field is -1, it should change to 0, right? Currently you have to click it once to turn it on (1) and again to turn it off (0). Maybe the code already handles -1 as false for those bool flags, but it'd put me more at ease if the value snapped to a known valid one.

NQNStudios:
Screenshot 2025-03-02 at 3 15 22 PM

Some special nodes are showing as "error" in the One-Time category.

src/doxy/mainpage.md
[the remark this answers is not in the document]
NQNStudios: Specifically on the mapgroup element.

> * `<mapgroup>` - Defines an LED group that won't be instanced by the tilemap – that is,
> there will only be one group, not one group in each cell of the tilemap. However,
> the LED controls defined inside the `<mapgroup>` _will_ be instanced, and will all
> be in the same group. So, it's an LED group that spans every cell of the tilemap.

I think my confusion here could be solved by editing what comes before this explanation.

> The `<tilemap>` tag represents a grid of identical elements based off the provided template.
It can contain any elements except for nested `<stack>`, `<pane>`, or `<tilemap>` elements.

If we append here Something like "Every child element of the tilemap will be instanced for *each* cell of the tilemap, unless it is a `<mapgroup>` element (see below)."

Then I would be prepared to understand what it means to be "instanced" vs not instanced.

> * `<mapgroup>` - Defines a combined LED group that will contain every instance of the LEDs it defines, which will be created for *each* cell of the tilemap.

NQNStudios:
When you scroll away from your selected SDF in the SDF picker, then scroll back returning it into view, it makes a sound when that LED becomes selected again. Ideally it should change state silently because it wasn't clicked again.

NQNStudios:
I think the SDF picker button is kind of missable where it is.

Screenshot 2025-03-02 at 3 40 34 PM

If the framed labels were narrowed you could fit it as a regular size button with a label on it, to the left of the Part A label. The Part A label is already wider than the Part B one, dunno if there's a reason for that.

CelticMinstrel:
> When an LED is placed next to ex2a, if the value in the field is -1, it should change to 0, right? Currently you have to click it once to turn it on (1) and again to turn it off (0). Maybe the code already handles -1 as false for those bool flags, but it'd put me more at ease if the value snapped to a known valid one.

I checked a lot of them in boe.specials.cpp – most of them just do a "greater than zero" check to delineate true from false. It might make sense to warp it to 0, but I decided that was a bit beyond the scope of what I was doing. It's probably something that would make sense for other things besides just toggle… though I'm not convinced it's a good idea in general, as it makes "selecting a different type" perform a possibly lossy change.

> Some special nodes are showing as "error" in the One-Time category.

Yes. They were originally "Unused", but I'd need to add an explicit node button dictionary entry to preserve that. I think in practice the solution is to squish them down so that there isn't a gap.

I'm pretty sure changing the internal number of a special node won't actually affect anything at all (feel free to correct me if I'm wrong) – the file references them by opcode name, and everything in the code uses the enum. As for legacy importing, it uses the original integer as the source node and the enum for the target, so it too wouldn't be affected. All I need to do is make sure that the squish happens in both the enum and the file with the list of all the opcode names, and it should be fine.

The other possibility is explicitly declaring them in the button dictionary. That would make them display as "Unused" like they used to.

> When you scroll away from your selected SDF in the SDF picker, then scroll back returning it into view, it makes a sound when that LED becomes selected again. Ideally it should change state silently because it wasn't clicked again.

Agreed. I'll see what I can do about that. Do you get the same in the editable string picker or is it _only_ the SDF picker that does this?

> I think the SDF picker button is kind of missable where it is.

Well, this isn't new (it's already the case for rectangle nodes that already had pickers before I did this).

PR #657, opened 2025-02-23
Replay fixes

### LINKED ISSUE #559: Record replays in memory
I think we could greatly streamline bug reporting in a few pretty simple steps:

1. Add a preference to "Record play session in memory" (call it something less technical than 'in memory' for clarity to players)
2. When the preference is enabled at launch-time, we set record to true but instead of a file, we write to a stream
2. When an error is caught at runtime, prompt the player that the replay can be saved to a file for uploading a reproducible report
3. Also provide a menu item for exporting the replay, in case of visual/other bugs which don't trigger error handling

NQNStudios:
A round of replay system fixes and improvements.
* replaying the debug help window was broken in 2 ways which I fixed
* Fix #559 -- When we observe bugs in random sessions, we no longer have to regret that we didn't pass `--record` at the command line. The replay data is kept in memory and writing it out can be enabled with debug action Z.
* Recording replays into the working directory by default is pretty wonky, because the working directory is usually a build output folder that might be clobbered by a `scons -c`, wiping out important replays. Also, if the working directory is the Visual Studio project folder, you end up adding untracked files to your repo folder which can get annoying. So - I've made the default replay directory a folder alongside the tempDir that already exists. On Windows, this is in AppData, on Mac it's in Application Support, etc. But if you pass an explicitly relative or absolute path, it is resolved as expected.
* I made a mistake a while ago that made the Debug action N (end scenario) not get recorded. Fixed!

### DIFF
--- a/src/fileio/fileio.cpp
+++ b/src/fileio/fileio.cpp
@@ -33,7 +33,7 @@ bool mac_is_intel(){
 	}
 	return _mac_is_intel;
 }
-fs::path progDir, tempDir, scenDir;
+fs::path progDir, tempDir, scenDir, replayDir;
 
 // This is here to avoid unnecessarily duplicating it in platform-specific files.
 cursor_type Cursor::current = sword_curs;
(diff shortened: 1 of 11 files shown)


NQNStudios:
I'm rounding up a bunch of these fixes as I've finally made a replay that contains (almost) all of the debug actions. More to come.

NQNStudios:
With those last two, the whole replay runs correctly.

PR #660, opened 2025-02-24
Retarget windows sdk in VS2017 project

NQNStudios:
These changes (aside from the gitignore one) are things that Visual Studio has to do automatically before I can build the code at all locally. They have to happen every time I switch branches, and often I have to stash save them to get git to let me to pull or do other things. Basically, if these don't break the CI, having them merged would make my life way easier.


NQNStudios:
I think the CI will work again if we bump the Visual Studio runners to Windows-2022 (which I have now done, so let's see what happens)

NQNStudios:
Squashed to one commit.

PR #662, opened 2025-02-24
Record scrollbar movement in dialogs

NQNStudios:
I forgot to stage one more replay system fix. This enables recording and replaying scrollbars in dialogxml scrollpanes.

### DIFF
--- a/src/dialogxml/dialogs/dialog.cpp
+++ b/src/dialogxml/dialogs/dialog.cpp
@@ -569,6 +569,16 @@ void cDialog::handle_events() {
 		}else if(replaying && has_next_action("field_selection")) {
 			cTextField& text_field = dynamic_cast<cTextField&>(getControl(currentFocus));
 			text_field.replay_selection(pop_next_action());
+		}else if(replaying && has_next_action("scrollbar_setPosition")){
+			Element& next_action = pop_next_action();
+			auto info = info_from_action(next_action);
+			std::string name = info["name"];
+			long newPos = boost::lexical_cast<long>(info["newPos"]);
+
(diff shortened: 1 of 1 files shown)


src/dialogxml/widgets/control.hpp
```diff
@@ -149,6 +149,7 @@ class cControl {
 	/// @throw xMissingAttr if a required attribute is missing
 	/// @throw xMissingElem if a required attribute is either missing or present in insufficient quantity
 	virtual void validatePostParse(ticpp::Element& who, std::string fname, const std::set<std::string>& attrs, const std::multiset<std::string>& nodes);
+	std::string name;
```
CelticMinstrel, line 152: I don't like the positioning of this data member in the class. I think it should be placed at the bottom of the class, together with all the other data members. Does it actually need to be protected? If so, I'd put a new protected section just above the private section at the bottom of the class.

CelticMinstrel:
I have two general comments on this.

1. If every control now stores its ID, I think there are a bunch of code paths that could be simplified, where the control map is searched to get the ID of self.
2. I don't like that the scroll pane now _requires_ an ID to function. I understand _why_ it does (controls that don't specify an explicit Id gets a randomized ID, which won't work with replays), but can we come up with a different way to solve this?

NQNStudios:
I think the ui's randomized ids are deterministic through a ui_rand object.

I added the id to this one because I had already recorded a replay where the id wasn't recorded, so I wanted to be able to edit the replay xml to retroactively set the name to something I knew it would be.

NQNStudios:
@CelticMinstrel I removed the unnecessary 'name' attribute from about-boe and confirmed that a scrollpane without a name will record and replay correctly.

PR #667, opened 2025-02-28
Debug actions work

### LINKED ISSUE #664: Debug: T to enter Town should move outdoor location
Debugging, T will put you in a town, but when you leave, you'll be at the old outdoor position. It's disorienting and also doesn't help if you were trying to get to the outdoor location of the town in question.

NQNStudios:
* Previewing dialogxmls, don't crash the game if the XML has an error in it (just show the message)
* Allow canceling the file picker for previewing dialogxmls
* Allow toggling debug mode in the startup menu. In the startup menu, you can only use actions that make sense.
* Debug enter town: new behavior - move the party's outdoors location to a town entrance if possible (Fix #664)
* Fix debug help menu crashing when trying to make '~' into a cKey
* Make the debug help window more readily expandible, and add more buttons.
    * Note: I still don't want to make it a stupidly tall dialog like the ones we've had to split up to fit on screens when scaled. It shouldn't get much bigger -- maybe it should paginate if it has to. Although, if it bleeds off the screen, at least it won't break players' default experience, because it's a debug feature
* New debug actions (useful from the startup menu): Shift-2, Shift-3, Shift-4 (@, #, $) launch into the core scenarios, skipping the intro dialog.
    * if the party is in a scenario already, confirm before switching

### DIFF
--- a/src/game/boe.actions.cpp
+++ b/src/game/boe.actions.cpp
@@ -2445,7 +2445,8 @@ void debug_launch_scen(std::string scen_name) {
 	}
 	// Take party out of scenario they're in
 	if(univ.party.is_in_scenario()){
-		// TODO: Confirm
+		if(cChoiceDlog("leave-scenario",{"okay","cancel"}).show() != "okay")
+			return;
 		handle_victory(true);
 	}
 	// Force party into the scenario, skipping the intro:
--- a/src/game/boe.actions.cpp
+++ b/src/game/boe.actions.cpp
@@ -2148,20 +2148,30 @@ void debug_magic_map() {
 	print_buf();
 }
 
+extern void outd_move_to_first_town_entrance(int town);
+
 void debug_enter_town() {
 	if(recording){
 		record_action("debug_enter_town", "");
 	}
+
+	std::vector<std::string> town_names;
+	for(cTown* town : univ.scenario.towns){
+		town_names.push_back(town->name);
+	}
+	int town = get_num_response(0, univ.scenario.towns.size() - 1, "Enter Town Number", town_names);
+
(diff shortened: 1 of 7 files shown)


src/game/boe.actions.cpp
```diff
@@ -2454,18 +2494,29 @@ void init_debug_actions() {
 	add_debug_action({'<'}, "Make one day pass", debug_increase_age);
 	add_debug_action({'>'}, "Reset towns (excludes the one you're in, if any)", debug_towns_forget);
 	add_debug_action({'!'}, "Toggle Special Node Step-through Mode", debug_step_through);
+
+	// Enter core scenarios, skipping intro:
+	add_debug_action({'@'}, "Quick-launch The Valley of Dying Things", []() -> void {debug_launch_scen("valleydy.boes");}, true);
+	add_debug_action({'#'}, "Quick-launch A Small Rebellion", []() -> void {debug_launch_scen("stealth.boes");}, true);
+	add_debug_action({'$'}, "Quick-launch The Za-Khazi Run", []() -> void {debug_launch_scen("zakhazi.boes");}, true);
```
CelticMinstrel, line 2501: The `->void` part is unnecessary. Actually, I think `()` can be omitted too but I don't really care one way or another on that.

CelticMinstrel:
> Note: I still don't want to make it a stupidly tall dialog like the ones we've had to split up to fit on screens when scaled.

There's always the option of making it two columns, or even using a scroll pane.

NQNStudios:
Oh, a scroll pane would be brilliant. Never would have thought of that, and I was just barely doing work on scrollbars.

I don't think we need it yet.

PR #676, opened 2025-03-03
use nested map for SDF names to save stack space. Fix #674

### LINKED ISSUE #674: Win-vs64 stack overflow.
I traced the stack overflow to its source, which is the declaration of `sdf_names` with 350x50 std::strings. When I change the dimensions to 50x50 it works.

So we probably need to use an expanding vector for this, or a map of maps instead.

### LINKED ISSUE #675: SDFs omitted from Scenedit docs TOC
I went to double-check whether the 350 is rows or columns, and expected the Scenedit documentation TOC https://openboe.com/docs/editor/Contents.html to have a heading for "SDF" or "Stuff Done Flags" because that's such a core concept. But there isn't one--I haven't found the info yet.

NQNStudios:
This fixes the stack overflow and clarifies the spatial nature of the SDF grid.

Fix #675

### DIFF
--- a/src/scenario/scenario.cpp
+++ b/src/scenario/scenario.cpp
@@ -605,3 +605,11 @@ std::string cScenario::get_feature_flag(std::string flag) {
 	if(iter == this->feature_flags.end()) return "";
 	return iter->second;
 }
+
+std::string cScenario::get_sdf_name(int row, int col) {
+	if(sdf_names.find(row) == sdf_names.end())
+		return "";
+	if(sdf_names[row].find(col) == sdf_names[row].end())
+		return "";
+	return sdf_names[row][col];
+}
--- a/src/scenario/scenario.cpp
+++ b/src/scenario/scenario.cpp
@@ -607,6 +607,9 @@ std::string cScenario::get_feature_flag(std::string flag) {
 }
 
 std::string cScenario::get_sdf_name(int row, int col) {
+	if(row < 0 || row >= SDF_ROWS || col < 0 || col >= SDF_COLUMNS){
+		throw "Tried to access SDF name for out-of-bounds flag (" + std::to_string(row) + ", " + std::to_string(col) + ")";
+	}
 	if(sdf_names.find(row) == sdf_names.end())
 		return "";
 	if(sdf_names[row].find(col) == sdf_names[row].end())
--- a/src/scenedit/scen.sdfpicker.cpp
+++ b/src/scenedit/scen.sdfpicker.cpp
@@ -46,19 +46,19 @@ bool cStuffDonePicker::handle_scroll(std::string item_hit) {
 	if(item_hit == "up") {
 		if(viewport.y > 0) viewport.y -= rows;
 	} else if(item_hit == "down") {
-		if(viewport.y < scenario.sdf_names[0].size() - rows) viewport.y += rows;
+		if(viewport.y < SDF_ROWS - rows) viewport.y += rows;
 	} else if(item_hit == "left") {
 		if(viewport.x > 0) viewport.x -= cols;
 	} else if(item_hit == "right") {
-		if(viewport.x < scenario.sdf_names.size() - cols) viewport.x += cols;
+		if(viewport.x < SDF_COLUMNS - cols) viewport.x += cols;
 	}
 	if(viewport.x == 0) dlog["left"].hide();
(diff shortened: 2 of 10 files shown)


NQNStudios:
Now I'm wondering whether saving out the scenario will end up accessing every flag name index and populating it with enough empty strings to get the stack overflow anyway.

NQNStudios:
I thiiiink I got to all of the code paths that would fill the map with empty data, so now a scenario with no sdf names won't use memory to store blank strings.

CelticMinstrel:
> populating it with enough empty strings to get the stack overflow anyway.

It wouldn't be a stack overflow even if it did – the map stores its data on the heap, not the stack.

CelticMinstrel:
> I think my change is in line with the way SDFs themselves were already stored as an array2d of char.

I'm not too sure what you mean by that, but… a 2-dimensional array doesn't specify which dimension is the rows and which is the columns. If anything, it's probably more common for the first index to be the row, because then it matches how you write out a 2-dimensional array literal in the source code.

Though it looks like my array2d definition suggests the opposite was intended. Is that what you meant? Still, that x and y doesn't really mean anything.

NQNStudios:
The documentation and the x and y on the typedef are consistent in implying the 350 is the columns in my opinion. But this is why I hate 2D arrays and if something is gonna reverse the coordinate order it should use r and c for the variable names imo.

We can switch all of it to make it the other way around if that works better with the picker and the Set Many SDFs node. I just think my reasoning was correct. :P

NQNStudios:
That makes perfect sense. The documentation could include screenshots of the sdf picker to show the spatial relationship.

NQNStudios:
Okay. There's no way we can change the actual shape of the array2d for SDFs without breaking a bazillion things. So the change has to be semantic: instead of (x,y) coordinates, SDF flags now use (r,c) coordinates. This is explained in the documentation. And I added a TOC element for SDFs.

src/scenedit/scen.sdfpicker.cpp
```diff
@@ -106,40 +106,43 @@ location cStuffDonePicker::run() {
 }
 
 void cStuffDonePicker::clamp_sdf() {
-	chosen_sdf.x = minmax(0, scenario.sdf_names.size() - 1, chosen_sdf.x);
-	chosen_sdf.y = minmax(0, scenario.sdf_names[0].size() - 1, chosen_sdf.y);
-	viewport.x = cols * floor(chosen_sdf.x / float(cols));
-	viewport.y = rows * floor(chosen_sdf.y / float(rows));
+	// Note: x and y in chosen_sdf are actually (r,c)
```
CelticMinstrel, line 109: This contradicts the other similar comments. Could that be the source of the crash you mentioned?

src/scenedit/scen.sdfpicker.hpp
```diff
@@ -16,6 +16,8 @@ class cTilemap;
 
 class cStuffDonePicker {
 	cDialog dlog;
+	// Note: x and y in initial_sdf and chosen_sdf are actually (c, r) and must be flipped when
```
CelticMinstrel, line 19: There's no harm in this comment. I think it's not really necessary, because "x is column and y is row" is just a simple mathematical fact. But I guess maybe it's not as obvious to some people.

src/scenario/scenario.cpp
```diff
@@ -605,3 +605,11 @@ std::string cScenario::get_feature_flag(std::string flag) {
 	if(iter == this->feature_flags.end()) return "";
 	return iter->second;
 }
+
+std::string cScenario::get_sdf_name(int row, int col) {
```
CelticMinstrel, line 609: Hmm… could also add a bounds check here, should we?

src/universe/party.hpp
```diff
@@ -37,8 +37,8 @@ namespace legacy {
 };
 
 // TODO: Duplicated in scenario.hpp
-template<typename T, size_t x, size_t y>
-using array2d = std::array<std::array<T, y>, x>;
+template<typename T, size_t rows, size_t cols>
```
CelticMinstrel, line 40: The comment is no longer true, right?

src/scenario/scenario.cpp
NQNStudios: Sure, I've done that.

PR #677, opened 2025-03-04
Move file-static RenderWindow and RenderTexture instances inside functions to fix GL race condition on startup

nater0000:
This change is not required on all platforms, but currently when compiling against SFML 2.6.2 static libraries, at runtime, the GL layer on Windows is encountering an Access Violation attempting to lock a critical section before main() is reached.


nater0000:
I did not mean to add the vcxproj or sln files. Should I close and re-open a new PR without those changes?

NQNStudios:
You can get the VS file changes out of it without closing this PR. From the command line of your local branch:

```
git reset --soft HEAD^     # This undoes the commit but leaves your changes staged
git restore --staged proj     # This unstages the changes to the project files
git commit -m "Move file-static RenderWindow/RenderTexture instances into functions

this fixes a GL race condition on startup" # This makes your commit again, with a shorter headline and more info split to lower lines, which will display better in the history
git push --force          # This rewrites the history of your branch, automatically updating the pull request without requiring you to make a new one
```

CelticMinstrel:
A couple of small amendments to NQN's suggestion.

Before running any of that, run the following command:

```bash
git show --stat HEAD
```

That will give you a line starting with "commit HASH", where HASH looks like a random string. Copy-past the hash. Then, instead of the commit commanf NQN suggested, use this commit one:

```bash
git commit -C HASH
```

Where HASH is the string you just copy-pasted.

nater0000:
I think it worked

src/game/boe.ui.cpp
```diff
@@ -21,8 +21,17 @@
...
+	{
+		static sf::RenderTexture instance;
+		return instance;
+	}
 	cToolbar toolbar;
 }
+sf::RenderTexture& cToolbar::cache()
+{
+	return UI::toolbar_gworld();
+}
```
CelticMinstrel, line 34: I don't think there's any need to have two different functions here. I'd recommend declaring `cToolbar::cache()` static in the header, and moving the cntent of `toolbar_gworld()` into it.

Also, I don't think it needs to be made public?

src/pcedit/pc.menu.cpp
```diff
@@ -4,7 +4,7 @@
 #include "pc.menus.hpp"
 
 OpenBoEPCEditMenu::OpenBoEPCEditMenu(sf::RenderWindow& window)
-	: mainPtr { window }
+	: mainPtr() { window }
```
CelticMinstrel, line 7: This is wrong. I'm pretty sure it won't even compile.

CelticMinstrel (changes requested):
This is actually pretty nice. It might be a huge change, but it's nice and simple – practically a find-and-replace for the most part. There are a couple of small issues, but once those are dealt with (and once I've worked out the CI problem in #674 / #675), I'll go ahead and merge it.

Thanks for the help!

CelticMinstrel:
Note: I meant #676 not #675.

src/pcedit/pc.menu.cpp
nater0000: Ah good catch! It looks like this class is only used in pc.menus.linux.cpp so I didn’t try compiling it on Win32.

src/game/boe.ui.cpp
nater0000: I did it this way just to be consistent with the other instances of sf::RenderWorld/Texture but please make any desired changes!

src/pcedit/pc.menu.cpp
NQNStudios: @CelticMinstrel I think the CI would have shown this error if you approve the workflow to run?

NQNStudios: I don't have a Linux setup right now so I can't check directly.

NQNStudios:
Closed in favor of #682

PR #683, opened 2025-03-06
On catching fatal error, prompt to save the log and report

NQNStudios:
This one probably needs some discussion and refinement.

The main idea is, if a player finds a crash in the game, I want to capture as much info about it as possible. So this prompts the player to save the replay log and links them to this issue tracker.

Drawback: GitHub doesn't support drag-and-drop for XML files. So players have to zip the XML file themselves. (I think we have a dependency that could do the zipping? But I don't know how to write that code.)

Drawback: GitHub requires an account before adding an issue. It might be more ideal to find an external bug tracker, or host our own portal for it, to remove friction for less technically inclined players.

If any of this sounds undesirable in the core repo, I can also have it be exclusive to Itch edition and point to my fork's issue tracker.

### DIFF
--- a/src/game/boe.main.cpp
+++ b/src/game/boe.main.cpp
@@ -201,6 +201,17 @@ eMenuChoice menuChoice=eMenuChoice::MENU_CHOICE_NONE;
 short menuChoiceId=-1;
 #endif
 
+static void handleFatalError(std::string what) {
+	showFatalError(what);
+	if(recording){
+		record_action("error", what);
+		extern fs::path log_file;
+		if(log_file.empty() && cChoiceDlog("ask-save-replay", {"yes", "no"}).show() == "yes") {
+			save_replay_log();
+		}
+	}
+}
+
 int main(int argc, char* argv[]) {
 #if 0
 	void debug_oldstructs();
@@ -233,13 +244,13 @@ int main(int argc, char* argv[]) {
 		close_program();
 		return 0;
 	} catch(std::exception& x) {
-		showFatalError(x.what());
+		handleFatalError(x.what());
 		throw;
 	} catch(std::string& x) {
-		showFatalError(x);
+		handleFatalError(x);
 		throw;
 	} catch(...) {
-		showFatalError("An unknown error occurred!");
+		handleFatalError("An unknown error occurred!");
 		throw;
 	}
 }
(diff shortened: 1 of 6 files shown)


CelticMinstrel:
> Drawback: GitHub doesn't support drag-and-drop for XML files. So players have to zip the XML file themselves. (I think we have a dependency that could do the zipping? But I don't know how to write that code.)

We have a GZIP library and an XML library, but there's a problem – they don't play well together. The GZIP library uses iostreams, but the XML library uses C streams. We could make them play well together if we edited the XML library though. (I've already made a few edits to it, and the library is no longer maintained, so I think this is a viable option.)

I do think I may have made a bad choice of XML library way back when I got this started, but it might be a bit late to change it now.

> Drawback: GitHub requires an account before adding an issue. It might be more ideal to find an external bug tracker, or host our own portal for it, to remove friction for less technically inclined players.

I do think we should get off of GitHub in general. If you have any particular suggestions for source code hosting, issue hosting, or even CI hosting, I'd like to hear them.

It probably would be really easy for me to host a bug tracker on openboe.com. I'm not so sure about hosting something like a GitLab or Gitea instance though – that seems to be tricky to set up. As for CI, I probably can't manage that easily on openboe.com.

> If any of this sounds undesirable in the core repo, I can also have it be exclusive to Itch edition and point to my fork's issue tracker.

Nah, this seems like a good idea to me. The one thing I'm slightly worried about though is, where is the replay log? Is it a global variable? Does any of this code use something that might have been clobbered when the stack was unwound? I think the chance of an issue is fairly low, but we should be careful. This also means that the things this accesses must be either global or owned by `main()`, declared outside the `try` block.

src/game/boe.main.cpp
```diff
@@ -227,19 +238,21 @@ int main(int argc, char* argv[]) {
 		menu_activate();
 		restore_cursor();
 
+		throw std::string { "fatal" };
```
CelticMinstrel, line 241: Obviously, this line will have to be removed before I can merge.

NQNStudios: Ahahahaha I can't believe I staged that

NQNStudios:
@CelticMinstrel Another thing we could do here is prompt if they'd like to make an autosave, if they're in a saveable game mode. To prevent loss of progress in case reloading would let them avoid triggering the crash.

NQNStudios:
> The one thing I'm slightly worried about though is, where is the replay log? Is it a global variable? Does any of this code use something that might have been clobbered when the stack was unwound? I think the chance of an issue is fairly low, but we should be careful. This also means that the things this accesses must be either global or owned by main(), declared outside the try block.

It's global in `replay.cpp`. But making sure this all works could be an opportunity to add a *really funny* debug action: Crash the game.

PR #689, opened 2025-03-08
2.0 Beta 1

NQNStudios:
This is my branch where new features won't be added, only bugfixes and optimizations in the core game.

@CelticMinstrel one thing so far that is worth noting: I added a background thread at game initialization, which we can use to preload things that will otherwise halt the game when requested. For now, this is specifically so that the `welcome` and `about-boe` dialogs can be shown instantaneously. I tried two different approaches to optimizing those dialogs, and they didn't work at all, so this is my best solution.


CelticMinstrel:
Expanding on the previous PR, I'm now adding even more pickers and also a way for the special node editor to be more dynamic.

Already done:

* Saved Item Rectangle editing was redone
* Signs and area rectangles can no longer be created from the list view
* Dialogue node editor has been upgraded with the new pickers
* New picker for job boards
* New variation of the skill picker for Has Enough of Statistic
* New visual spell pattern picker

Remaining to-do list:

* [X] Fix all cases where the special node field prompt overflows the available space, and cases where the prompt is the field name when it should be "Unused".
* [ ] Add help strings for every special node.
* [X] Use Personality picker for Start Conversation
* [X] Picture and pattern picker should take a click on the picture as equivalent to a click on the LED.
* [X] Variable Town Entry editor needs some work.
* [ ] Visual picker for tiny icons in the terrain editor.
* [X] Use the SDF picker in many more places.
* [ ] Maybe add minor variations of the location picker for rectangles, paths, or other cases where multiple locations are needed. Details still uncertain.
* [ ] Add a picker for the strings in the Story Dialog special node.
* [ ] Add pickers (or special options in existing pickers) for pointers. Details still TBD.

NQNStudios:
It's really bad for me--so much so that bumping it to the background during the splash screen doesn't actually hide the whole process. As intended, when the splash screen would normally end, `join()` is called to make sure the loading finished--and for me this adds a splash screen delay of several seconds every time. I'm still looking for ways to make the dialogs actually initialize faster, because it's so annoying.

The two things I tried before pre-constructing the dialogs entirely were:

* Pre-cacheing the frames of all the controls, so relative positioning code could be skipped. https://github.com/nqnstudios/cboe/commit/24ed495c50a9112c5db5b713207bf41723086a5f
   * Maybe I did something wrong so it was doing the positioning code regardless. But this didn't speed things up at all, and broke the layout (definitely my fault) -- but I concluded that the relative positioning isn't the actual slow part.
   * And then I felt dumb for not formally profiling the process before attempting optimizing... something I still haven't done because I don't know how.
* Pre-caching the DialogDefn of the dialog
    * What this revealed to me is that it's definitely the construction of the dialog that the problem comes from

src/gfx/render_text.hpp
```diff
@@ -63,6 +64,6 @@ void clear_scale_aware_text(sf::RenderTexture& texture);
 void win_draw_string(sf::RenderTarget& dest_window,rectangle dest_rect,std::string str,eTextMode mode,TextStyle style,bool right_align = false);
 void win_draw_string(sf::RenderTarget& dest_window,rectangle dest_rect,std::string str,eTextMode mode,TextStyle style, break_info_t break_info,bool right_align = false);
 break_info_t calculate_line_wrapping(rectangle dest_rect, std::string str, TextStyle style);
-size_t string_length(std::string str, TextStyle style, short* height = nullptr);
+size_t string_length(std::string str, TextStyle& style, short* height = nullptr);
```
CelticMinstrel, line 67: The style should be a const reference. The measurement cache can be declared mutable.

CelticMinstrel:
> Pre-cacheing the frames of all the controls, so relative positioning code could be skipped.

This sounds like it should work. I don't know what you might have done wrong in your attempt. One thing I notice is that you serialized the frames as just an array, but it seems plausible that what happened is the order of the controls isn't the same on each launch, due to randomized IDs. I think applying the cached frames in `cControl::parse` rather than after all controls have been loaded would likely be less problematic. Then it only depends on the order of the XML, which is fixed.

It would probably also be a good idea to store the control IDs in the serialized layout file. Then you can print a warning if the ID of the frame doesn't match the ID of the control it's being applied to. There'd be a lot of false positives in a naive implementation of that (due to randomized IDs) but that can probably be dealt with somehow.

(Note: In case it's not obvious, ID is the name attribute on the control.)

CelticMinstrel:
This is a great place for RAII.

```cpp
struct profiler {
	profiler(std::string context) : context(context) {
		start_profile(context);
	}
	~profiled() {
		end_profile(context);
	}
	std::string context;
};
```

NQNStudios:
> This is a great place for RAII.
> 
> ```c++
> struct profiler {
> 	profiler(std::string context) : context(context) {
> 		start_profile(context);
> 	}
> 	~profiler() {
> 		end_profile(context);
> 	}
> 	std::string context;
> };
> ```

Well, the idea of it is that you can put START_PROFILE_FUNC; at the start of all the functions you're interested in, END_PROFILE_FUNC before they return. And there is a stack of the call contexts that you're in--so a subcall takes over the timer while that function is running, but returns the timer to the calling function when finished.

NQNStudios:
Oh, if the stack was static, probably it could.

CelticMinstrel:
I'm a bit unsure as to what you're confused about. What RAII brings in is just that you can put a profiler at the top of a function and you won't miss the end context call because the function returned early. Instead of needing to put `END_PROFILE_FUNC` before every return statement, all you need is a `PROFILE_FUNC` at the top of the function and those end calls are added automatically – and not just when the function returns but also when it throws an exception, or when a function it _called_ throws an exception that it doesn't catch.

CelticMinstrel:
Yeah, the code I posted wasn't meant to replace what you wrote but merely wrap it, hence why I wrote it as calling the functions you defined. (Though you could in theory make that constructor and destructor _be_ the start and end profile functions, but that doesn't really make a difference.)

NQNStudios:
I didn't have enough time this morning to fully locate the problem. Here's example output from profiling with the directives I added in the WIP commit:

```
about-boe
----
about-boe: 8.05423
parseContent: 0.000307
parseAttribute: 4.2e-05
calculate_layout: 4e-05
validatePostParse: 2.6e-05
recalcRect: 2.1e-05
parse: 1.6e-05
parseChildControl: 5e-06
loadFromFile: 3e-06
----
welcome
----
welcome: 2.92701
parseAttribute: 5.7e-05
calculate_layout: 3.9e-05
recalcRect: 2.5e-05
parseContent: 2.3e-05
validatePostParse: 1.6e-05
parse: 1.3e-05
loadFromFile: 1e-06
----
```

What this means, if my profiling code is correct, is that the about-boe dialog is taking 8 seconds of time that *isn't* in any of the functions where I added `PROFILE_FUNC`--and I feel like I added it to most of the major suspects....

NQNStudios:
It looks like win-scons is defining __func__ as empty...
> build\obj\dialogxml\dialogs\dialog.cpp(154): error C2512: 'profiler': no appropriate default constructor available

NQNStudios:
Boost provides a CURRENT_FUNCTION definition that gives more detailed info than __func__ anyway:

```
about-boe
----
about-boe: 9.7326
virtual bool cTextMsg::parseContent(ticpp::Node &, int, std::string, std::string, std::string &): 0.00028
virtual bool cControl::parseAttribute(ticpp::Attribute &, std::string, std::string): 5.5e-05
void cTextMsg::calculate_layout(): 5e-05
virtual void cControl::validatePostParse(ticpp::Element &, std::string, const std::set<std::string> &, const std::multiset<std::string> &): 3.1e-05
virtual void cTextMsg::recalcRect(): 2.6e-05
virtual void cControl::parse(ticpp::Element &, std::string): 1.9e-05
bool cContainer::parseChildControl(ticpp::Element &, std::map<std::string, cControl *> &, std::string &, std::string): 6e-06
void cDialog::loadFromFile(const DialogDefn &): 3e-06
void cDialog::recalcRect(): 3e-06
----
welcome
----
welcome: 2.88705
virtual bool cControl::parseAttribute(ticpp::Attribute &, std::string, std::string): 4.4e-05
void cTextMsg::calculate_layout(): 3.7e-05
virtual void cTextMsg::recalcRect(): 2.4e-05
virtual void cControl::validatePostParse(ticpp::Element &, std::string, const std::set<std::string> &, const std::multiset<std::string> &): 1.9e-05
virtual bool cTextMsg::parseContent(ticpp::Node &, int, std::string, std::string, std::string &): 1.8e-05
virtual void cControl::parse(ticpp::Element &, std::string): 1.5e-05
void cDialog::recalcRect(): 2e-06
void cDialog::loadFromFile(const DialogDefn &): 1e-06
```

src/fileio/fileio_party.cpp
```diff
@@ -382,60 +382,62 @@ bool load_party_v2(fs::path file_to_load, cUniverse& real_univ, bool manual){
...
-			std::istream& fin = partyIn.getFile("save/setup.dat");
-			if(!fin) {
-				if(manual) showError("Loading Blades of Exile save file failed.");
-				return false;
-			}
-			file.readFrom(fin);
-			auto& page = file[0];
-			for(size_t i = 0; i < univ.party.setup.size(); i++) {
-				page[std::to_string(i)].extract(univ.party.setup[i]);
+		if(!preview){
```
CelticMinstrel, line 392: I would like it if you avoided adding a level of indentation here. It does mean redundantly checking `if(preview)` a few times instead of just once, but the optimizer should take care of that.

src/fileio/fileio_party.cpp
```diff
@@ -382,60 +382,62 @@ bool load_party_v2(fs::path file_to_load, cUniverse& real_univ, bool manual){
...
-		std::istream& fin = partyIn.getFile("save/out.txt");
-		if(!fin) {
-			if(manual) showError("Loading Blades of Exile save file failed.");
-			return false;
-		}
-		univ.out.readFrom(fin);
+			// Load outdoors data
+			std::istream& fin = partyIn.getFile("save/out.txt");
+			if(!fin) {
+				if(!preview) showError("Loading Blades of Exile save file failed.");
```
CelticMinstrel, line 429: It looks like this check is now inside a larger `if(preview)` block, making it redundant.

CelticMinstrel:
> Now those dialogs open fast enough that preloading is not necessary, so I reverted that stuff, and the calls to PROFILE_FUNC which I used to diagnose the problem.

Suggestion: It may be worthwhile to undo the revert and instead define `PROFILE_FUNC` so that it does nothing if `ENABLE_PROFILING` is not defined. Most builds will not define it, and then you can define it at a later time if you need to profile.

It looks like your updated lambda relies on the assumption that it will be called for consecutive indices starting from zero. That's fine as long as it's true (and I'm pretty sure it is), but I think you should add a comment about that as well.

Also, please don't forget my comment earlier about the making a paramater const.

src/fileio/fileio_party.cpp
NQNStudios: I can also get rid of the extra indentation by bringing the loading of exported graphics *up*, and then checking if(preview) for an early return.

This should be fine as long as there's not a dependency between loading exported graphics and the things that I'm skipping. Which is something I'm already assuming, by skipping those things for the preview load.

Issue #67, opened 2018-02-15
There is a white line around the bottom and right sides of the Save First button when quitting the Character Editor.


daerogami:
This is an item logged from the To-Do. If you are familiar with or can recreate this bug, please provide details here.

NQNStudios:
This is referring to the intentional highlighting of the default button. 

<img width="461" alt="Image" src="https://github.com/user-attachments/assets/3e08ac7f-7ad0-4a61-9794-3f6da50c3fa2" />

On some buttons, one or more sides of the white border get covered up so it looks buggy.

NQNStudios:
Ah -- the inset visual effect just looks worse on dark backgrounds. I wonder how it can be made better?

CelticMinstrel:
I don't think that would help in the current implementation. It's just an outline, not a filled rect or anything.

PR #697, opened 2025-03-16
2.0 Beta 1 Pt 2

NQNStudios:


src/game/boe.items.cpp
```diff
@@ -894,31 +894,46 @@ short get_num_response(short min, short max, std::string prompt, std::vector<std
...
 	numPanel.run();
 	
 	return numPanel.getResult<int>();
 }
 
 static bool select_pc_event_filter (cDialog& me, std::string item_hit, eKeyMod) {
 	me.toast(true);
-	if(item_hit != "cancel") {
+	if(item_hit == "pick-all"){
+		me.setResult<short>(7);
```
CelticMinstrel, line 916: This doesn't seem great. 6 is used nearly universally to mean "the entire party", and _only_ here that meaning is conveyed by 7 instead?

src/game/boe.graphutil.cpp
```diff
@@ -372,6 +372,26 @@ void draw_fields(location where){
...
+		Draw_Some_Item(fields_gworld,calc_rect(4,3),terrain_screen_gworld(),where_draw,1,0);
+	if(univ.town.is_ash(where.x,where.y))
+		Draw_Some_Item(fields_gworld,calc_rect(5,3),terrain_screen_gworld(),where_draw,1,0);
+	if(univ.town.is_bones(where.x,where.y))
+		Draw_Some_Item(fields_gworld,calc_rect(6,3),terrain_screen_gworld(),where_draw,1,0);
+	if(univ.town.is_rubble(where.x,where.y))
+		Draw_Some_Item(fields_gworld,calc_rect(7,3),terrain_screen_gworld(),where_draw,1,0);
+
+	// In space/foreground:
+	// TODO some of these could still be reordered, maybe.
```
CelticMinstrel, line 394: I looked through them. The only other one that seems like it could be moved to the bottom is special spot, though that could just as easily make sense to move to the _top_.

But aside from that, there are definitely some oddities in the ordering. I would expect crate, barrel, and block to basically be on the same "layer" – generally you only have one of the three on a tile. Probably it makes sense to draw the more "permanent" ones first:

* barrel
* crate
* block
* web
* fire and force barrier (the animated ones)
* quickfire
* bottom half of force cage (note: the top half is drawn in a completely different place, after characters)

Then the more transient ones on top:

* Walls (force, ice, blade, fire)
* Antimagic field
* Clouds (stinking, sleep)

The one exception to this is the special spot, where it makes sense for it to be visible through anything and thus go on the top… though I'm not sure if it _should_. Maybe it's better for it to go at the bottom.

src/game/boe.newgraph.cpp
```diff
@@ -1039,16 +1041,22 @@ short scan_for_response(const char *str) {
...
-void handle_target_mode(eGameMode target_mode, int range) {
+void handle_target_mode(eGameMode target_mode, int range, eSpell spell) {
 	overall_mode = target_mode;
 	// Lock on to enemies in range: 
 	if(has_feature_flag("target-lock", "V1") && get_bool_pref("TargetLock", true)){
+		// Skip this for spells that don't target enemies
+		switch(spell){
+			case eSpell::DISPEL_SQUARE: case eSpell::DISPEL_BARRIER: case eSpell::DISPEL_SPHERE:
+				return;
+		}
```
CelticMinstrel, line 1052: I think you've missed some here?

The most obvious is `DISPEL_FIELD`, since the three you _do_ have are also dispels. It's an item-only spell, maybe that's why you missed it.

Then there's unlock and the two versions of move mountains… or are they excluded because they can only be cast in town and this path is only used in combat? _Is_ this only used in combat?

Then there's the summoning spells (of which there are 12 – Sticks to Snakes and Demon are the less obvious ones), which definitely don't target monsters. I don't think it necessarily makes sense to put the summons amongst the enemies, especially since summoning is quite short range. There's also the new Flash Step I added as an item spell.

The field spells are debatable, but I guess you'd normally place them so that enemies are caught in them, so it's fine to have the auto-framing behaviour.

NQNStudios: Having recorded a replay with every priest spell yesterday, I can confidently say the code path isn't used for peace-mode priest spells. And that the priest summoning spells don't make you target, they position the summoned creatures arbitrarily.

I should re-check the mage summons because I do think some of them are targeted.

src/game/boe.items.cpp
NQNStudios: I think it would make sense for cancel to be -1 and all to be 6.

I'd just have to carefully change callers that expect 6 to be a cancellation.

src/game/boe.newgraph.cpp
CelticMinstrel: > And that the priest summoning spells don't make you target, they position the summoned creatures arbitrarily.

Incorrect. They _do_ position the summoned creatures arbitrarily in town mode, but let you select their positions if you cast in combat.

NQNStudios: Oh, ok. Well since there's no harm in adding cases here that might not ever be hit, I can just add *all* the summoning spells, unlock, move mountains, just cast a wide net.

NQNStudios: Or we could add something to the spell info struct for this.

src/game/boe.items.cpp
NQNStudios: @CelticMinstrel I set about trying to change this, and I what I found is that the magic numbers and comparisons of return values are so pervasive in the code it feels impossible to change every instance without breaking something. Return values from select_pc() get stored in global state which ends up getting checked in other areas of the code. Lots of variables that get assigned to select_pc() are expecting 6 as their empty value, and getting assigned 6 from other places.

I don't think it's a worthwhile refactor to make.

CelticMinstrel: For what it's worth, there are _definitely_ game mechanics that target the entire party, rather than an as-yet-unspecified member of the party.

src/game/boe.newgraph.cpp
CelticMinstrel: Just in case: Strengthen Target definitely targets enemies. It's used by a cursed wand that makes your enemy stronger.

I think all the others should be pretty obvious.

NQNStudios:
But not ones where it’s a choice between 1 and all, no? I can only think of
the “____ All” spells and there’s no selection dialog for those.

src/game/boe.items.cpp
NQNStudios: But not ones where it’s a choice between 1 and all, no? I can only think of
the “____ All” spells and there’s no selection dialog for those.

src/game/boe.items.hpp
```diff
@@ -37,7 +37,22 @@ void reset_item_max();
...
+	ONLY_LIVING,
+	ONLY_LIVING_WITH_ITEM_SLOT,
+	// Slightly different from the previous: this now checks not just for a slot,
+	// but also for carrying capacity, and will stack charges even if all slots are full.
+	// When you use this, you have to set the global item_store to your item, first.
+	ONLY_CAN_GIVE,
+	// Same as the previous, but in combat, only show *adjacent* living PCs who can take the item.
+	ONLY_CAN_GIVE_FROM_ACTIVE,
+	ONLY_DEAD,
+};
```
CelticMinstrel, line 55: This was originally `enum class eSelectPC { ONLY_LIVING, ANY, ONLY_DEAD };`, right? Reordering it probably  breaks the Select Target special node.

NQNStudios: That didn't need to be fixed, because select target nodes already translated between their own integer mapping and the call args to `select_pc`:

https://github.com/NQNStudios/cboe/blob/99eb9962b5063a8bcfbe9adf2b2d4a9874ff6c65/src/game/boe.specials.cpp#L2701

There's no casting integers to the enum type.

src/dialogxml/dialogs/strchoice.cpp
```diff
@@ -12,10 +12,15 @@
...
 #include <boost/lexical_cast.hpp>
+#include <boost/algorithm/string/case_conv.hpp>
 
 #include "dialogxml/widgets/field.hpp"
 #include "fileio/resmgr/res_dialog.hpp"
 #include "sounds.hpp"
+#include "gfx/render_shapes.hpp"
+
+const sf::Color HILITE_COLOUR = Colours::LIGHT_GREEN;
+const sf::Color NORMAL_COLOUR = Colours::WHITE;
```
CelticMinstrel, line 23: This is incorrect. The normal colour is not white. It is white in the game and PC editor and black in the scenario editor. I'm pretty sure I have a `defClr` or similar somewhere that handles this.

src/dialogxml/dialogs/strchoice.cpp
```diff
@@ -156,6 +162,88 @@ bool cStringChoice::onOkay(cDialog& me){
...
+		if(highlightSearch()) found_next = true;
+	}
+
+	while(!found_next){
+		if(page == loop_from_page){
+			// Looped already, found nothing. Avoid infinite loop:
+			if(looped_once){
+				break;
+			}
+			output.setText("Reached the " + loop_from_str + ". Starting from the " + loop_to_str);
```
CelticMinstrel, line 200: I know we don't have localization yet, and plenty of stuff is in an unlocalizable state, but it would be nice if we can avoid adding _more_ stuff that's unlocalizable. In this case, the thing to avoid is subsentence string concatenation. I don't think it's difficult to simply set `loop_from_str` and `loop_to_str` to just contain the whole sentence? Then this would become `output.setText(loop_from_str + " " + loop_to_str")`.

src/dialogxml/dialogs/strchoice.cpp
NQNStudios:
```
	if(!attrs.count("color") && !attrs.count("colour") && getDialog()->getBg() == cDialog::BG_DARK)
		setColour(sf::Color::White);
```
Is it this, in message.cpp?

NQNStudios: oh `parent.getDefTextClr()`

CelticMinstrel:
Very low priority, but I'd suggest visually distinguishing the escape button in some way, similar to how the default button is drawn outlined.

Something to open an issue for after merging if you don't think it's worth doing in here.

NQNStudios:
@CelticMinstrel Now that the changes are really starting to stack up (100 commits! maybe my biggest PR?), it might be nice to do some merging and start anew. The other two PRs have taken commits from this one, so I'd like to see them merged and then rebase this one.

NQNStudios:
> Very low priority, but I'd suggest visually distinguishing the escape button in some way, similar to how the default button is drawn outlined.
> 
> Something to open an issue for after merging if you don't think it's worth doing in here.
> 

I agree there should be a visual indicator. While we're on that subject, I also don't know if you saw my comment on #67 that I think the existing visual highlight for the default button looks bad on certain backgrounds causing it to be perceived as a bug. I actually think it would be nicer to have color-tinted button sprites, like the ok buttons could be blueish and the cancel buttons could be more reddish.

CelticMinstrel:
> Now that the changes are really starting to stack up (100 commits! maybe my biggest PR?), it might be nice to do some merging and start anew.

Sounds good, but I notice that you have a bunch of fixup commits in here. Did you plan to rebase to squash those?

Also, while I'm here, I think it makes sense to map Ctrl+F to `key_find` rather than `key_none` (though `key_none` can still exist to catch random modifier key combinations). I don't quite remember how the special keys work though so I'm not sure if that affects the XML-set key mapping.

NQNStudios:
> Sounds good, but I notice that you have a bunch of fixup commits in here. Did you plan to rebase to squash those?

Yeah! If you merge the other two PRs I'll be rebasing this one onto master, and I'll autosquash when I do that.

> Also, while I'm here, I think it makes sense to map Ctrl+F to key_find rather than key_none (though key_none can still exist to catch random modifier key combinations). I don't quite remember how the special keys work though so I'm not sure if that affects the XML-set key mapping.

I see there's not a key_find in eSpecKey. Would I add that? And I'm currently not sure where key combos get mapped to eSpecKeys although I can probably find it. (If I already modified that part of the code in this PR I have forgotten. 😵‍💫 )

CelticMinstrel:
> If you merge the other two PRs

I don't understand. There are no other PRs opened against master at the moment.

Certainly I could merge the tutorial and bolts PRs, but they are not against master so I don't understand why they would be treated as a blocker for this.

NQNStudios:
Ok, yeah, I was thinking about things in a convoluted way that would have involved merging the tutorial PR *then* merging the tutorial branch to master *then* merging this PR to master. It can be way simpler.

I should autosquash this, then it can merge to master. Then the upstream tutorial branch should be synchronized with master. Then I rebase the tutorial PR without the cherry-picks which will be made unnecessary by merging this.

NQNStudios:
I've autosquashed the fixups.

The bolt PR onto scen_update might as well merge so it's one less thing to think about.

If you merge this and rebase the tutorial branch, let me know so I can rebase the tutorial PR. That one could be merged once rebased, but doesn't really need to be unless you are interested in starting designing the actual scenario before I might get to it.

CelticMinstrel:
https://github.com/calref/cboe/pull/697#discussion_r2003960215

> But not ones where it’s a choice between 1 and all, no? I can only think of
the “____ All” spells and there’s no selection dialog for those.

I think I completely lost track of this thread. I'm not aware of any cases where the _player_ needs to choose between targeting one character or all characters, but the Affect category of special nodes are set to work on either one or all.

This is just a comment, not a request for changes or anything (it's a bit late for that now).

CelticMinstrel:
Sorry for the delay on merging this. I'll try to get to it this weekend.

NQNStudios:
No worries. I had to wrap up another project and I'm applying to jobs so this hasn't been a blocker or anything.

Issue #362, opened 2024-06-12
Windows NSIS installer generation not working


NQNStudios:
In PR #359 for win-scons I wasn't able to get the installer to generate, but I want to make this a separate issue so I can work on other things for a while.

PR #708, opened 2025-03-22
Tutorial base

NQNStudios:
I cherry-picked some commits from #697 which are required for launching the scenario and conveniently launching it from the editor.

I added the tutorial scenario, unpacked, with nothing in it (except one special spot & node). The only town is large, because I'm imagining it as the "hub" where the party enters different portals to the modular lesson room towns, and we have a lot of sections to cover so I don't want to run out of space in the hub town. (Although we could split it...) In the next PR I'll lay out a bit of the hub to show how I'm envisioning it will accommodate a linear but modular flow through lessons where we can add lessons in the middle without town layout headaches.

All of the build options should package the tutorial (though I've only tested scons on Mac, I followed the patterns for how the other projects did the other scenarios.)

I cherry-picked another commit (the cControl.replaceText() one) to let me reuse a confirmation dialog for if the player has a party loaded before clicking the tutorial button.

The logic of the original confirmation I was reusing, left a possibility of wiping player progress without confirmation, so I fixed that.

Thought I had to change the dialog schema (I did not), but was having such a hard time searching for the dialog documentation file that I added a note in dialog.xsd.

The tutorial now only contains a special node that ends itself. `handle_victory()` checks if the tutorial you're completing is the tutorial, and if it is, it asks whether to start VoDT.

(This involved adding the last yes/no confirmation dialog I intend on making a new dialogxml for.)


NQNStudios:
Hmmm I see I made it so if you debug quit the tutorial it will still prompt about VoDT. I'll fix that a bit later.

src/game/boe.actions.cpp
```diff
@@ -3406,13 +3413,19 @@ void new_party() {
 	if(recording){
 		record_action("new_party", "");
 	}
-	if(overall_mode != MODE_STARTUP) {
-		std::string choice = cChoiceDlog("restart-game",{"okay","cancel"}).show();
+	if(party_in_memory) {
+		cChoiceDlog confirm("restart-game",{"okay","cancel"});
+		(confirm.operator->())->getControl("warning").replaceText("{{action}}", "Starting over");
```
CelticMinstrel, line 3418: Huh? Wait. Isn't this `(confirm.operator->())->` just the same as `confirm->`? There's no reason to manually call the arrow operator here as far as I know…

rsrc/dialogs/restart-game.xml
```diff
@@ -3,8 +3,8 @@
 <?xml-stylesheet href="dialog.xsl" type="text/xsl"?>
 <dialog defbtn='cancel'>
 	<pict type='dlog' num='23' top='9' left='9'/>
-	<text top='6' left='53' width='249' height='31'>
-		Starting over will delete the current party.
+	<text name='warning' top='6' left='53' width='249' height='31'>
+		{{action}} will unload the current party without saving.
```
CelticMinstrel, line 7: I actually don't like this phrasing. I can sort of understand what was bad about the old phrasing, but this might be even worse in different ways. My main gripe is that "unload" feels a bit too technical. I think it should be phrased more along the lines of "You will lose unsaved changes if". Maybe not _exactly_ like that but closer to that.

Side note: This is also unlocalizable. That's not a simple fix (unlike the last time I mentioned localizability) so I'll let it pass for now.

CelticMinstrel:
> The only town is large, because I'm imagining it as the "hub" where the party enters different portals to the modular lesson room towns, and we have a lot of sections to cover so I don't want to run out of space in the hub town.

This strikes me as quite weird. If the tutorial is going to be linear, why would there be a hub? I thought the sort of "modular" we discussed was more of a "we can easily insert another module into the flow" rather than "players can choose which parts they play through".

> `handle_victory()` checks if the tutorial you're completing is the tutorial, and if it is, it asks whether to start VoDT.

I don't really like this. I think it makes more sense to dump them out to the main menu so they can pick a scenario. But I guess it's not that bad as long as that's what happens when they choose "no".

> All of the build options should package the tutorial (though I've only tested scons on Mac, I followed the patterns for how the other projects did the other scenarios.)

I can see at least one you missed – the Windows installer. There are no changes to the NSIS scripts in this pull request.

src/game/boe.actions.cpp
NQNStudios: Yeah, this was very silly of me

NQNStudios:
> This strikes me as quite weird. If the tutorial is going to be linear, why would there be a hub?

I keep using poor choice of words for the concepts in my head. But what I was thinking of as a "hub", actually I don't think is required at all.

> I thought the sort of "modular" we discussed was more of a "we can easily insert another module into the flow" rather than "players can choose which parts they play through".

I think it should be both?

> I don't really like this. I think it makes more sense to dump them out to the main menu so they can pick a scenario. But I guess it's not that bad as long as that's what happens when they choose "no".

That is what happens if you choose no.

> I can see at least one you missed – the Windows installer. There are no changes to the NSIS scripts in this pull request.

I didn't realize those did something with packing the scenarios, because it doesn't work at all (#362).

It's something that runs after the normal Visual Studio build, right? So build-scen will already have been called, I just need to add it to the copy step, looks like. Though the installer still won't work.

NQNStudios:
I've rebased and updated all this -- it will come again in a later PR.

PR #722, opened 2025-04-10
2.0 Beta 1 Part 3

### LINKED ISSUE #61: Targeting reticle doesn't always draw properly
Win10, commit cb9f0ed

Sometimes it works, sometimes i get this 
badreticle

I suspect this has to do with the hidden-from-view behavior or bad masking with certain tiles. Need to confirm.

Did further testing. This is a bug with the new UI scaling feature. The targeting mask/boundary looks like its dimensions are not being resized as well.

### LINKED ISSUE #721: Rotated wall spells don't always put the wall where the targeting reticle shows
Sometimes the wall appears off-by-one from where the framed tiles appear when I'm targeting. This can be frustrating.

NQNStudios:
This was all innocent enough minor fixes until the 3 commits before the last one. Then things got weird.

As I'll be pushing more later let me specify the commits that were major weird stuff...

[Remember window positions as preferences](https://github.com/calref/cboe/pull/722/commits/04c6b6a164413f72237303c12187346baa938a60)
[Let the minimap be hidden by other applications in focus](https://github.com/calref/cboe/pull/722/commits/334d6875fda7d293410ba519f2d1be11d9d76427)
[WIP standardize mouse position translation](https://github.com/calref/cboe/pull/722/commits/a83ef7a2b1491d09460e7f0d4ec3aa68a7459331) -- The idea of this one was to fix #721 while also making future misalignments between visual selection and actual selection less likely to happen. I have not tested enough to be confident that it *did* fix the bug.

Fix #61


NQNStudios:
I'm deep in territory I don't wanna be in. Very yikes.

src/location.cpp
```diff
@@ -132,12 +132,12 @@ rectangle_size_delegate& rectangle_size_delegate::operator-=(int val) {
 	return *this;
 }
 
-rectangle_size_delegate& rectangle_size_delegate::operator*=(int val) {
+rectangle_size_delegate& rectangle_size_delegate::operator*=(float val) {
```
CelticMinstrel, line 135: Could you make these `double` instead of `float`?

src/game/boe.text.cpp
```diff
@@ -326,13 +326,25 @@ void put_item_screen(eItemWinMode screen_num) {
...
-						if(item.charges > 0 && item.ability != eItemAbil::MESSAGE && (stat_screen_mode == MODE_INVEN || stat_screen_mode == MODE_SHOP))
-							sout << '(' << int(item.charges) << ')';
+					else
+						sout << item.name << "  ";
+
+					// Charges:
+					bool show_charges = item.max_charges > 0;
+					// Show charges for unidentified ammunition, but not other items
+					if(item.missile <= 0)
+						show_charges &= item.ident;
```
CelticMinstrel, line 339: Isn't this the missile graphic number? Is that really a good metric…?

src/game/boe.actions.cpp
```diff
@@ -1402,6 +1473,30 @@ void update_item_stats_area(bool& need_reprint) {
 	need_reprint = true;
 }
 
+location mouse_window_coords() {
+	location where_curs = sf::Mouse::getPosition(mainPtr());
+	where_curs = mainPtr().mapPixelToCoords(where_curs, mainView);
```
CelticMinstrel, line 1478: Did you test this on Mac? It looks like the kind of code that might need the "input monitoring" permission, but I'm not entirely sure. Maybe it's more lenient with mouse than keyboard.

src/dialogxml/dialogs/dialog.cpp
```diff
@@ -43,6 +43,8 @@ const short cDialog::BG_DARK = 5, cDialog::BG_LIGHT = 16;
 short cDialog::defaultBackground = cDialog::BG_DARK;
 cDialog* cDialog::topWindow = nullptr;
 void (*cDialog::redraw_everything)() = nullptr;
+std::function<sf::RenderWindow&()> cDialog::get_mini_map;
+bool* cDialog::map_visible_p = nullptr;
```
CelticMinstrel, line 47: I _really_ don't like all this minimap handling in `cDialog`. It has no business knowing that a minimap even exists.

src/game/boe.actions.cpp
NQNStudios: Yeah, I wrote it on mac, but I did have the input permissions enabled. So just now I turned off the input permissions to see if it still worked, and it does, so I think it should be fine?

src/dialogxml/dialogs/dialog.cpp
NQNStudios: I agree, and I think I've come up with a better way. Instead of have a get_mini_map() delegate it should pass an onLostFocus and onGainedFocus delegate, so the game code can supply all of that logic.

NQNStudios:
I removed the functions I removed because I grepped and saw nowhere that would have triggered them being called, either by calling them directly or by setting bools that would result in it happening. That said, I didn't thoroughly look into what those functions were *supposed* to do, and so it bears looking closer to be more than 90% sure I didn't remove anything that would be good to keep.

CelticMinstrel:
I think most special nodes would be previewable, so maybe the preview flag should default to true and be explicitly disabled via `.no_preview()` on nodes that can't be previewed.

NQNStudios:
> I think most special nodes would be previewable, so maybe the preview flag should default to true and be explicitly disabled via `.no_preview()` on nodes that can't be previewed.

Well, right now each type of node needs to have its preview manually implemented, so inverting that flag would just make the button appear for a bunch of node types that aren't actually previewable until that work gets done. I might be splitting this work into a separate PR and not implementing every type of preview right away.

CelticMinstrel:
> Well, right now each type of node needs to have its preview manually implemented, so inverting that flag would just make the button appear for a bunch of node types that aren't actually previewable until that work gets done.

There is a sensible default for a node that doesn't specify a preview method, though – use msg1 and msg2 for a 2-string dialog.

NQNStudios:
Good idea, I'll put that in and see how many cases it adequately covers.

CelticMinstrel:
2f2f61b6d3dd6a59cf9f2a16e175e683837c8d5c probably should not have changed the filesize, yet the size is more than double. Can you explain that? The files should be saved without needless metadata and in indexed 8-bit format.

NQNStudios:
> 2f2f61b6d3dd6a59cf9f2a16e175e683837c8d5c probably should not have changed the filesize, yet the size is more than double. Can you explain that? The files should be saved without needless metadata and in indexed 8-bit format.
> 

That will be for the same reason it happened last time--default GIMP PNG export settings. How did you end up exporting it to get the size down again last time?

I'll fix it soon but another change is coming to the same file, I have someone working on a more visible baby hydra.

CelticMinstrel:
> How did you end up exporting it to get the size down again last time?

I think I unticked every box in the export settings dialog, but I don't remember if I did anything else in addition to that.

NQNStudios:
[efab3ee](https://github.com/calref/cboe/pull/722/commits/efab3ee34937efab60e96ab52a03fb3baf08969e) I fixed the size inflation of monst10.png and added a darker baby hydra I had a friend make. We agree it could be prettier but so could all the other sprites--this at least fixes the visibility.

CelticMinstrel:
Re: 344a5983dff614d5cd27ff2e90f517c8591f8621

I'm pretty sure I fixed this already…? What's going on here? But 12 _is_ the correct top coordinate when I look at the graphics sheet…

Re: e9bdb9f6d259229061b02d6654e3e7dece0b5afb

Uhh, re-initializing the rects doesn't really seem like the correct approach… I think I'd prefer to have two different rects initialized and select the correct one based on context.

Re: 86bcf2e69faf028852416a91e7aaf49f20d80787

Can we use `\u2026` (ie `…`) instead of `...`?

Re: 264c806440ec0134217b2af8a6cd994b5c537a11

This seems questionable. Changing where just one terrain special is checked seems like it could affect a whole bunch of things that assumed it would be checked in the old place.

First of all, I think we should check what happens in classic BoE in this situation, and stick with that.

Secondly, from my understanding, your bug was that a node that's supposed to run after unlocking the door also runs if you fail to unlock, or choose not to try. That'd be easy to fix in the scenario, by adding an "if terrain this type" node in front of the chain.

Re: d7476326a61c8b825b9dc4074749395abb38fa31

What exactly is more understandable about removing "forced"? I'm not saying it definitely isn't more understandable, but I just don't understand your reasoning here.

This is basically what used to be the "Secret Passage" special node in classic BoE. That is, it allows you to pass through a terrain that is set as impassable.


Re: bd168c9cde19295b8659f49684ebe5456d27f4fa

It looks like your `case TALK_BACK` wrongly falls through to `case TALK_ASK` if the feature flag is not set or if the dialog history is empty. Also, your two labelled sections at the bottom of the function are indented one level too far.

NQNStudios:
> I'm pretty sure I fixed this already…? What's going on here? But 12 is the correct top coordinate when I look at the graphics sheet…

The same source rectangle was hard-coded wrong in more than one place.

NQNStudios:
> First of all, I think we should check what happens in classic BoE in this situation, and stick with that.

I just checked, and the behavior that I went with implementing is what happens in classic BoE.

I'll see if I can keep that behavior while not messing so much with the order of operations.

> What exactly is more understandable about removing "forced"? I'm not saying it definitely isn't more understandable, but I just don't understand your reasoning here.

What gets me about "force allow" is that it feels like an oxymoron to me, especially when the node itself is called "Prevent Action."

NQNStudios:
> It looks like your case TALK_BACK wrongly falls through to case TALK_ASK if the feature flag is not set or if the dialog history is empty. Also, your two labelled sections at the bottom of the function are indented one level too far.

You're right. I've just fixed this with a `git commit --amend`.

> Can we use \u2026 (ie …) instead of ...?

Unfortunately that character renders like this:

Screenshot 2025-05-07 at 11 17 01 AM

NQNStudios:
> Uhh, re-initializing the rects doesn't really seem like the correct approach… I think I'd prefer to have two different rects initialized and select the correct one based on context.

In the case of init_shopping_rects() which now responds to changes in the string length of the item cost, re-initializing *is* required--so it seems fine to me to do it for this one too. I mean, maybe we would want to re-initialize these ones as well, for example, to make it so each rectangle only covers the actual rendered text and not all the extra space. (I don't think that would be better, but it's a plausible example because we do that with LED labels.)

I thought about doing it the other way and felt like that would entail making item_buttons a 3-dimensional array... `[which_item][which_button][which_mode]` and that seemed pretty silly.

NQNStudios:
> > This seems questionable. Changing where just one terrain special is checked seems like it could affect a whole bunch of things that assumed it would be checked in the old place.
> >
> >First of all, I think we should check what happens in classic BoE in this situation, and stick with that.
>
>I just checked, and the behavior that I went with implementing is what happens in classic BoE (node doesn't happen until you STEP ONTO the space, AFTER it is unlocked).
>
>I'll see if I can keep that behavior while not messing so much with the order of operations.

Now that I think about it, it feels potentially wrong that special spots would run before special terrain handling in any case whatsoever, not just locked doors. I think I'm going to double-check with the original source release, and maybe test more cases in legacy BoE.

NQNStudios:
I think I figured it out. Just need to test something else in classic BoE.

src/game/boe.dlgutil.cpp
```diff
@@ -1170,13 +1219,25 @@ void handle_talk_node(int which_talk_entry) {
...
+			place_talk_str(save_talk_str1,save_talk_str2,0,dummy_rect);
+			reset_talk_words();
+			return;
+	}
+
+	FINISH_TALK_NODE:
+		dialog_history.push_back({can_save_talk, false, -1, {save_talk_str1, save_talk_str2}});
+		place_talk_str(save_talk_str1,save_talk_str2,0,dummy_rect);
+		reset_talk_words();
+		return;
```
CelticMinstrel, line 1240: This stuff is still indented one level too deep (except for `if(false){` and `}`).

NQNStudios: whoops, sorry.

CelticMinstrel:
I don't like this: 1aeac0c1631106d6f07c02cb5fe71cffedf3fc43

If you want an acid damage type, then add an acid damage type. Don't just half add it. And why on earth is it -1? That makes no sense. Also, if you're adding it then it needs its own boom.

Personally, I don't think we really need a separate acid damage type. I don't especially mind having one either, though.

NQNStudios:
> And why on earth is it -1? That makes no sense.
The next ones say 'keep these two last'--I am afraid of changing their values (or any of the values in this enum) because magic number references to them could be anywhere.

NQNStudios:
I'm still processing it as magic damage so as not to change the game balance--magic resistance has always applied to it, and there are no features anywhere that would add acid resistance. The reason for "half-adding" it now, is mainly to use the intended sound effect and make it easy to add a graphic for it later.

CelticMinstrel:
Well, this commit changed SPECIAL from 9 to 8: d877df5101e6601d9158e8ba90aad5b56e91da8a

So that suggests to me that changing it back to 9 shouldn't cause any problems.

CelticMinstrel:
> I'm still processing it as magic damage so as not to change the game balance--magic resistance has always applied to it, and there are no features anywhere that would add acid resistance.

Its existence in the enum means you can set acid resistance and whatever other things you can come up with.

I think the correct approach is to handle it like I handled the new races (Vahnatai, Goblin, Skeletal Undead, possibly a few others) – basically, anything that checks for magical also covers acid. So, magic resistance also gives acid resistance, and protection from magic also protects from acid, etc.

CelticMinstrel:
Re: 1d3a641ef04904c440a5278ca1cd399905fa8652

For this you probably need to update unit tests and maybe some of the sample scenario files in the test folder.

NQNStudios:
For a temporary acid boom, I started with this monster graphic from BoA:

image

I isolated the slime, then scaled it down to fit BoE boom size.

BoA credits two artists, one for the line art, and one for the icons. Andrew Hunter is the icon artist, and already credited as the main artist for BoE. I think somewhere you said we already have Andrew Hunter's permission to use additional graphics he made for Spiderweb games? Anyway, we can always ask to make sure.

CelticMinstrel:
> For a temporary acid boom, I started with this monster graphic from BoA:

Uhh… what? I think using graphics from BoA is fairly questionable – I even left out a bunch of graphics that Mistb0rn created specifically for BoE because of that. Besides that, the graphic doesn't even match the aesthetics of the other booms. That makes it a really bad fit even as a temp graphic.

Maaaybe you could get something useful out of that slime if you mirrored it vertically and squished it in just the right way (not uniformly – the protrusions would need to shrink more than the centre), but I'd much rather take one of the pre-existing booms on that sheet and modify it.

> BoA credits two artists, one for the line art, and one for the icons. Andrew Hunter is the icon artist, and already credited as the main artist for BoE. I think somewhere you said we already have Andrew Hunter's permission to use additional graphics he made for Spiderweb games? Anyway, we can always ask to make sure.

We have Andrew Hunter's permission to use graphics he created for the Exile trilogy. I don't know if that permission extends to graphics created for the Avernum series. I didn't even know he had created those graphics.

NQNStudios:
I reeeeaaaally don't think it sticks out so horribly as you're saying. Just a reminder that I'm doing my best and trying to make as many improvements as I can over here with limited time to work. I tried fiddling with the pre-existing booms and couldn't come up with anything that felt distinct enough. I do think what I came up with is decent.

NQNStudios:
Also, damaging terrains write out their damage type in terrain.xml as an integer, so making changes to eDamageType values could definitely be destructive.

CelticMinstrel:
> Also, damaging terrains write out their damage type in terrain.xml as an integer, so making changes to eDamageType values could definitely be destructive.

That should be fixed, but there's no need to worry about it here – there won't be any damaging terrains that do special damage in any scenario.

> Resist Magic used to not help with magic damage!

I think the point of resist magic was improved saving throws, not reduced damage. (Assuming you mean the status effect.) That said, I don't have any problem with changing that, especially considering it _does_ reduce fire and ice damage.

> `add_string_to_buf("  It's acid!"); // Maybe?`

How about "It burns!"? Or something along those lines.

> I reeeeaaaally don't think it sticks out so horribly as you're saying. Just a reminder that I'm doing my best and trying to make as many improvements as I can over here with limited time to work. I tried fiddling with the pre-existing booms and couldn't come up with anything that felt distinct enough. I do think what I came up with is decent.

I don't really know if it could be called decent, but I do know that it doesn't really fit in with the others and I don't really like it.

> I just feel some of these reactions lately imply what I did is utterly baffling--in an iterative process where things aren't finished, no less.

I truly _was_ baffled when you picked something from Blades of Avernum… I can't think of anything else that counted as truly baffling though.

src/fileio/fileio_scen.cpp
```diff
@@ -1066,6 +1070,63 @@ void readScenarioFromXml(ticpp::Document&& data, cScenario& scenario) {
...
+			editor_state.last_out_edited = readLocFromXml(*elem);
+		}else if(type == "last-town") {
+			elem->GetText(&editor_state.last_town_edited);
+		}else if(type == "drawing"){
+			editor_state.drawing = str_to_bool(elem->GetText());
+		}else if(type == "editing-town"){
+			editor_state.editing_town = str_to_bool(elem->GetText());
+		}else if(type == "town-view-state"){
+			Element* child = elem->FirstChildElement();
+			while(child != nullptr){
```
CelticMinstrel, line 1100: Please use an `Iterator<Element>` for this.

src/fileio/fileio_scen.cpp
```diff
@@ -1066,6 +1070,63 @@ void readScenarioFromXml(ticpp::Document&& data, cScenario& scenario) {
...
+					center = readLocFromXml(*child);
+				}else if(child_type == "viewing-mode"){
+					child->GetText(&viewing_mode);
+				}
+				child = child->NextSiblingElement(false);
+			}
+			editor_state.town_view_state[num] = {center, viewing_mode};
+		}else if(type == "out-view-state"){
+			Element* child = elem->FirstChildElement();
+			while(child != nullptr){
```
CelticMinstrel, line 1114: Please use an `Iterator<Element>` for this.

src/fileio/fileio_scen.cpp
```diff
@@ -1066,6 +1070,63 @@ void readScenarioFromXml(ticpp::Document&& data, cScenario& scenario) {
...
+				child->GetValue(&child_type);
+				if(child_type == "section"){
+					section = readLocFromXml(*child);
```
CelticMinstrel, line 1117: Please put this information as x and y attributes on the `<out-view-state>` tag.

src/fileio/fileio_scen.cpp
```diff
@@ -1066,6 +1070,63 @@ void readScenarioFromXml(ticpp::Document&& data, cScenario& scenario) {
...
+		}else if(type == "drawing"){
+			editor_state.drawing = str_to_bool(elem->GetText());
+		}else if(type == "editing-town"){
+			editor_state.editing_town = str_to_bool(elem->GetText());
+		}else if(type == "town-view-state"){
+			Element* child = elem->FirstChildElement();
+			while(child != nullptr){
+				child->GetValue(&child_type);
+				if(child_type == "num"){
+					child->GetText(&num);
```
CelticMinstrel, line 1103: Please put this information as an attribute on the `<town-view-state>` tag.

src/game/boe.specials.cpp
```diff
@@ -4150,7 +4156,7 @@ void townmode_spec(const runtime_state& ctx) {
 				showError("Invalid attitude (0-Friendly Docile, 1-Hostile A, 2-Friendly Will Fight, 3-Hostile B).");
 				break;
 			}
-			univ.town.monst[spec.ex1a].attitude = eAttitude(spec.ex1b);
+			univ.town.monst.at(spec.ex1a).attitude = eAttitude(spec.ex1b);
```
CelticMinstrel, line 4159: Using `at` here doesn't help anything. It still crashes, just reliably (ie, every time) and sooner (ie, not later on because of data corruption). What's really needed is a manual range check… which means you might as well stick to the `[]` operator.

src/fileio/fileio_scen.cpp
NQNStudios: When I set it up that way,

```
Iterator<Element> child;              
for(child = child.begin(elem->FirstChildElement()); child != child.end(); child++){
```

The loop body never executes. I swear I'm following the exact same pattern as the enclosing loop, but it doesn't work, begin() returns end() even when elem->FirstChildElement() returns the proper result. When I switched to the pointer way it did work, and I just did not care to debug a non-trivial tinyxml bug when I can't find documentation of how the iterator is supposed to work.

NQNStudios: Ohhhh. I think I see the problem, and tutorial_ticpp.txt actually does have info on this. Another thing I'd like to make easier for new contributors to find and learn from.

NQNStudios: It has to be like this:

```
			for(child = child.begin(&(*elem)); child != child.end(); child++){
```
Dereferencing the iterator to get the Element, then passing its address to child.begin(). What misled me was that the enclosing loop is calling FirstChildElement() on the document, but when doing that on the inner loop it actually tries to iterate the inner child. This all makes sense to me now.

NQNStudios: or instead of `&(*`, `.Get()`

CelticMinstrel:
Yeah, the ticpp iterator is a bit weird, but there are plenty of working examples in the source code already.

PR #764, opened 2025-06-24
Custom Graphics Tools

NQNStudios:
I think this is pretty nifty.

* You can specify an image editor in the preferences, then launch it from the custom graphics window editing the current sheet.
* There is a reload button for capturing your external changes back into the interface/storing an undo action for those edits. ~~(This isn't yet handled perfectly for sheets not contiguous with #0).~~
* Select a range of tiles in the current sheet for modification ~~(still need to show them highlighted)~~.
* Import other graphic icons on top of them
* Specify what graphic is the background, and the editor will do its best to strip it out for you
* Specify a new background, and the editor will overlay it for you.

(In other words, a possible solution for #554 )

This video shows me transposing cave walls onto grass:
https://youtu.be/qvR4MGFmUMI


NQNStudios:
Notes to self:

- frame custom dialog pic halves
	- frame custom talk pic halves
	- import monster graphic
	- import tall monster graphic
	- import wide monster graphic 
	- import big monster graphic
	- icon tool buttons re-frame the selection to icons that will be changed
		- bools for tall selection shape?
	- all importing/framing classify?

PR #770, opened 2025-06-26
Fix scenario picker, update README & build script

### LINKED ISSUE #768: Scenario picker broke
What on earth happened here?

All I did was alt-click on Start Scenario and it's completely broken. Though, clicking one of those three slots _does_ open the scenario that _should_ be shown there.

### LINKED ISSUE #769: Windows build depends on 7zip - README doesn't mention this.
I swear I remember that this has come up before-- but I can't find an open issue about it. On a fresh Windows laptop I ran this:

> The command "C:\Users\Nat\repos\cboe\proj\vs2017\..\..\pkg\win\build-scen.bat C:\Users\Nat\repos\cboe\proj\vs2017\..\..\rsrc\bases\bladbase C:\Users\Nat\repos\cboe\proj\vs2017\..\..\rsrc\bases\bladbase.boes" exited with code 255.

The command, when run directly, fails because 7zip is not installed on my computer. 

7zip is used [here](https://github.com/NQNStudios/cboe/blob/6da30745984f39115eb4cdbacc1ab5cfc2da9f5a/pkg/win/build-scen.bat#L2) -- This is fine, as long as there's not a built-in alternative.

I just need to add the dependency to the README.

NQNStudios:
Fix #768 Fix #769

### DIFF
--- a/pkg/win/build-scen.bat
+++ b/pkg/win/build-scen.bat
@@ -1,5 +1,6 @@
 
 set Zip="C:\Program Files\7-Zip\7z.exe"
+%Zip% || @echo 7-zip must be installed first. && exit /B
 
 @echo Packing scenario %2...
 if exist %2 del /F %2
--- a/pkg/win/build-scen.bat
+++ b/pkg/win/build-scen.bat
@@ -1,6 +1,6 @@
 
 set Zip="C:\Program Files\7-Zip\7z.exe"
-%Zip% || @echo 7-zip must be installed first. && exit /B
+%Zip% >NUL || @echo 7-zip must be installed first. && exit /B
 
 @echo Packing scenario %2...
 if exist %2 del /F %2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,7 +125,7 @@
 			]
 		},
 		win-scons: {
-			runs-on: windows-2019,
+			runs-on: windows-2022,
 			env: {
 				VCPKG_BINARY_SOURCES: "clear;x-gha,readwrite"
 			},
@@ -164,7 +164,7 @@
 			]
 		},
 		# win-mingw: {
-		# 	runs-on: windows-2019,
+		# 	runs-on: windows-2022,
 		# 	steps: [
 		# 		{
 		# 			name: checkout,
--- a/.github/workflows/scripts/win/scons-build.bat
+++ b/.github/workflows/scripts/win/scons-build.bat
@@ -10,6 +10,6 @@ REM  @echo "%%i"
 REM  call "%%i" x86_amd64
 REM )
 
-call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64
+call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64
 
 scons bits=64 %*
--- a/README.md
+++ b/README.md
@@ -85,6 +85,8 @@ For Linux builds, the following additional dependencies are required:
   - or, if cmake is available when you call `scons`, TGUI will be built from source automatically
 - zenity command-line tools
 
+For Windows builds, [7-zip](https://7-zip.org/download.html) is required.
+
 If you are using the Visual Studio toolset, we recommend installing
 [vcpkg](https://github.com/Microsoft/vcpkg) to manage these dependencies.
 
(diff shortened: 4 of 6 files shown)


CelticMinstrel:
Is that second commit correct? It looks like it launches 7-Zip and prints an error if that fails. I don't know exactly what happens when 7-Zip launches – does it open a UI, print a banner message, or what?

NQNStudios:
Yeah, your interpretation of the script change is correct. It works as I expected for both cases (7-zip exists vs. 7-zip does not exist). No UI opens if it exists, this is the printout:

```
C:\Users\Nat\repos\cboe>"C:\Program Files\7-Zip\7z.exe"

7-Zip 24.09 (x64) : Copyright (c) 1999-2024 Igor Pavlov : 2024-11-29

Usage: 7z <command> [<switches>...] <archive_name> [<file_names>...] [@listfile]

<Commands>
  a : Add files to archive
  b : Benchmark
  d : Delete files from archive
  e : Extract files from archive (without using directory names)
  h : Calculate hash values for files
  i : Show information about supported formats
  l : List contents of archive
  rn : Rename files in archive
  t : Test integrity of archive
  u : Update files to archive
  x : eXtract files with full paths

<Switches>
  -- : Stop switches and @listfile parsing
  -ai[r[-|0]][m[-|2]][w[-]]{@listfile|!wildcard} : Include archives
  -ax[r[-|0]][m[-|2]][w[-]]{@listfile|!wildcard} : eXclude archives
  -ao{a|s|t|u} : set Overwrite mode
  -an : disable archive_name field
  -bb[0-3] : set output log level
  -bd : disable progress indicator
  -bs{o|e|p}{0|1|2} : set output stream for output/error/progress line
  -bt : show execution time statistics
  -i[r[-|0]][m[-|2]][w[-]]{@listfile|!wildcard} : Include filenames
  -m{Parameters} : set compression Method
    -mmt[N] : set number of CPU threads
    -mx[N] : set compression level: -mx1 (fastest) ... -mx9 (ultra)
  -o{Directory} : set Output directory
  -p{Password} : set Password
  -r[-|0] : Recurse subdirectories for name search
  -sa{a|e|s} : set Archive name mode
  -scc{UTF-8|WIN|DOS} : set charset for console input/output
  -scs{UTF-8|UTF-16LE|UTF-16BE|WIN|DOS|{id}} : set charset for list files
  -scrc[CRC32|CRC64|SHA256|SHA1|XXH64|*] : set hash function for x, e, h commands
  -sdel : delete files after compression
  -seml[.] : send archive by email
  -sfx[{name}] : Create SFX archive
  -si[{name}] : read data from stdin
  -slp : set Large Pages mode
  -slt : show technical information for l (List) command
  -snh : store hard links as links
  -snl : store symbolic links as links
  -sni : store NT security information
  -sns[-] : store NTFS alternate streams
  -so : write data to stdout
  -spd : disable wildcard matching for file names
  -spe : eliminate duplication of root folder for extract command
  -spf[2] : use fully qualified file paths
  -ssc[-] : set sensitive case mode
  -sse : stop archive creating, if it can't open some input file
  -ssp : do not change Last Access Time of source files while archiving
  -ssw : compress shared files
  -stl : set archive timestamp from the most recently modified file
  -stm{HexMask} : set CPU thread affinity mask (hexadecimal number)
  -stx{Type} : exclude archive type
  -t{Type} : Set type of archive
  -u[-][p#][q#][r#][x#][y#][z#][!newArchiveName] : Update options
  -v{Size}[b|k|m|g] : Create volumes
  -w[{path}] : assign Work directory. Empty path means a temporary directory
  -x[r[-|0]][m[-|2]][w[-]]{@listfile|!wildcard} : eXclude filenames
  -y : assume Yes on all queries
```

I guess that's pretty annoying if you're running build-scen.bat directly from the command line, but in the use case I tested (building through visual studio), I didn't see or notice it in the build output, which is generally less obtrusive/more ephemeral because the IDE aggregates errors for you in a UI anyway. Though I can't say for sure it didn't print and I'm trying to verify that.

Apparently 7-zip doesn't have a "quiet flag"--weird.

Not sure if win-scons builds will pipe all that into the terminal.

pkg/win/build-scen.bat
```diff
@@ -1,5 +1,6 @@
 
 set Zip="C:\Program Files\7-Zip\7z.exe"
+%Zip% || @echo 7-zip must be installed first. && exit /B
```
CelticMinstrel, line 3: > I guess that's pretty annoying if you're running build-scen.bat directly from the command line

Looks like that annoyance could be easily avoided by just redirecting to `NUL`.


```suggestion
%Zip% >NUL || @echo 7-zip must be installed first. && exit /B
```

PR #773, opened 2025-08-01
Scenario picker fixes

NQNStudios:
I threw the whole scenario archive in my scenarios folder and found some bugs.

* Some scenarios like 100 Evil Sliths don't start with a letter and the alphabetical buttons don't handle that (it will crash)
* Scenarios with unrecognized special nodes will show an error even if you're just loading the header for the scenario picker

This fixes them and adds a '#' button which pages the picker to the first page of custom scenarios.

### DIFF
--- a/rsrc/dialogs/pick-scenario.xml
+++ b/rsrc/dialogs/pick-scenario.xml
@@ -75,6 +75,7 @@
 	<button name='x' def-key='x' type='tiny' relative='pos pos-in' rel-anchor='prev' top='0' left='4'>X</button>
 	<button name='y' def-key='y' type='tiny' relative='pos pos-in' rel-anchor='prev' top='0' left='4'>Y</button>
 	<button name='z' def-key='z' type='tiny' relative='pos pos-in' rel-anchor='prev' top='0' left='4'>Z</button>
+	<button name='#' def-key='shift 3' type='tiny' relative='neg pos-in' anchor='j' top='0' left='32'>#</button>
 
 	<button name='cancel' type='regular' top='330' left='5'>Cancel</button>
 	<button name='prev' type='left' def-key='left' relative='pos pos-in' rel-anchor='prev' top='0' left='14'/>
--- a/src/fileio/fileio_scen.cpp
+++ b/src/fileio/fileio_scen.cpp
@@ -266,8 +266,11 @@ bool load_scenario_v1(fs::path file_to_load, cScenario& scenario, eLoadScenario 
 		return false;
 	}
 	port_item_list(&item_data);
-	scenario.import_legacy(temp_scenario);
-	scenario.import_legacy(item_data);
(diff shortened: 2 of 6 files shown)


CelticMinstrel:
> Scenarios with unrecognized special nodes will show an error even if you're just loading the header for the scenario picker

How can a legacy scenario have an unrecognized special node?

NQNStudios:
Not too difficult, no.

When I say 'the whole scenario archive', I'm referring to [this one](https://spiderwebforums.ipbhost.com/topic/33604-scenario-archive-super-editor-and-boe-on-linux/) arranged by The Lurker via Kelandon's archive and other sources. I've copied it to a repository [here](https://github.com/NQNStudios/cboe-scenarios) in which I ran a script to extract every zip file, because each scenario was in a separate one. (I'm experimenting with the possibility of shipping Itch Edition with the complete archive, which would obviate the need for an online scenario browser unless/until we start seeing new scenario releases again.)

Printing the identity of the scenarios that contain unrecognized nodes, it turns out there's only one.

```
Unrecognized node type found: 115
Unrecognized node type found: 232
Unrecognized node type found: 33
Unrecognized node type found: 160
scenario.scen_name: 
file_to_load: "/Users/nat/Library/Application Support/Blades of Exile/Scenarios/BoEArchFull/Normal/Suspension of Tranquility, The/Suspension of Tranquility/tranquil.exs"
```

Its name field is blank but there's the filename and node numbers.

CelticMinstrel:
> I've copied it to a repository here in which I ran a script to extract every zip file, because each scenario was in a separate one. 

Did you extract the zips on a Mac? Did you manually convert all the .meg files to data-fork files?

> Printing the identity of the scenarios that contain unrecognized nodes, it turns out there's only one.
> Its name field is blank but there's the filename and node numbers.
> EDIT: Wait a sec, I've found more, but I think they only show the error when you enter the scenario (probably because the unrecognized nodes are on a town or outdoor node, not scenario node?)

Okay, comparing those numbers to [here](https://github.com/calref/cboe/blob/legacy-win32/game/classes/consts.h#L707) doesn't reveal any matches, so I think there's a good chance that this just comes from garbage data in the file. Not sure how that could've happened though…

NQNStudios:
> Did you extract the zips on a Mac?

Yes.

> Did you manually convert all the .meg files to data-fork files?

No, should I? How do I do that?

NQNStudios:
To be honest I don't even know what a resource file vs resource fork vs data fork is. This stuff is from a few years before the first time I had my own computer.

That StackOverflow gives this one-liner, but it's unclear what to replace with what when invoking it...

`cp resource_file/..namedfork/rsrc  resource_fork`

CelticMinstrel:
Something like this, maybe?

```
mv something.meg something.meg.temp
cp something.meg.temp/..namedfork/rsrc something.meg
```

NQNStudios:
Closing in favor of #788

Issue #779, opened 2025-08-04
highlight friendly and charmed NPCs?


NQNStudios:
I think it's too easy to cast a spell or ranged attack on a friendly (charmed monsters included) and make the whole town hate you.

NQNStudios:
<img width="424" height="529" alt="Image" src="https://github.com/user-attachments/assets/88adbc88-4cf5-4945-b4fe-50821c8f67f4" />

Maybe the enemy/ally highlighting should happen at all times during combat mode, not just targeting.

CelticMinstrel:
> Maybe the enemy/ally highlighting should happen at all times during combat mode, not just targeting.

Not sure if I like the "grid" look that this creates, but the outline around friendly/hostile monsters seems fine. How does it handle larger monsters? I'd say the ideal situation is one outline around the whole monster, rather than separately outlining each space it occupies, but would that be too difficult?

Definitely this hilite shouldn't happen outside of combat mode. I'm not 100% sure what I think of it even in targeting mode; not really sure if it should happen outside of target mode. (Do remember also that target mode can happen outside of combat.)

This could be a preference, but I'm not sure if that's a good idea either.

Issue #727, opened 2025-05-05
Map overhaul


NQNStudios:
Based on what I was thinking yesterday, there are some things I thought of that would make the automap much more useful, especially outdoors. I think (possibly as a feature flag legacy scenarios won't have, and with preferences to disable the new features), it should allow for certain tiles and/or regions to be highlighted once they appear on your map, and show a tooltip when you mouse over them.

- [ ] Town entrances
- [ ] Horses/Boats you own
- [ ] Signs 
- [ ] Town areas

Also, I do think I want the map to show PC positions during combat.

PR #780, opened 2025-08-07
Many QOL features and bug fixes

NQNStudios:
My free time is gonna be a lot more limited in the near future so I wanted to make a big push and implement every quality of life feature left on my checklist, even if it meant moving too fast and getting messy.

I have now done that. (As well as fix a lot of legacy scenario compatibility stuff, sometimes in ways that presuppose the existence of my scenario archive submodule in the build and `meta.xml` files which currently only contain encodings. It also assumes that legacy scenarios in the custom scenario folder are in individual folders, not dumped into a flat folder structure with lots of floating READMEs.)

So not all of this should be merged upstream, but most of it should be, once sorted out, vetted and made more sane. I figured I would make one big PR so you can see everything I've done, but I will split it into smaller PRs and/or cherry-picks like we did with the big Fosnola PRs.

Implements #779 

Implements #727 - my minimap overhaul lets you see sign text, area descriptions, vehicle locations, and town entrances by mousing over them. I think this is a huge QOL improvement, but may spoil certain things if the player reads a sign's tooltip but the sign is inaccesible (ie on the other side of a wall). I think I'm comfortable with that tradeoff. It could also be made a preference, "Annotated minimap" or something. I don't think it should be a feature flag, because it improves the experience in legacy scenarios just as much as new ones.

I think this contains everything in PR #773, and probably more related fixes that should be added to that PR.


CelticMinstrel:
> In master, it's not actually on the first page. Each page has 3 slots, and vestigial code is in place in preparation for adding an extra slot to put Busywork on the first page. It actually cannot be accessed at all, in the current master branch.

Ah, I see. I think someone (probably ADoS) said we shouldn't put it on the first page until it has undergone some revisions to meet a minimum quality standard, but certainly it's bad if it can be accessed at all before then.

> It's just not worth it to me to over-engineer systems for on-the-fly support.

What overengineering? You already wrote code to do that work for the purpose of producing the meta file. Just leave the code (or a minor variation on it) active in the released build?

NQNStudios:
I've moved off of this branch -- closing for now.

